PR #1, opened 2021-04-20
Example showing usage of resource from an external ROS package in Drake

azeey:
Reopening https://github.com/RobotLocomotion/drake-external-examples/pull/194 here.

This adds an example of loading a URDF from an external ROS package. The URDF contains a mesh with a `package://` type URI. For this to be found, I populate the package map from `AMENT_PREFIX_PATH` as described in #170.

In addition to the pendulum related packages, I copied the `scripts` directory from  https://github.com/RobotLocomotion/drake-external-examples for the convenient installation scripts that were referenced in the README.

cc @IanTheEngineer , @EricCousineau-TRI

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/1)

PR #2, opened 2021-04-20
[DO NOT MERGE] Add generic systems for Drake ROS integration

hidmic:
Opening PR for awareness and discussion only. This patch builds on top of [sloretz/drake_ros2_demos#generic_systems](https://github.com/sloretz/drake_ros2_demos/tree/generic_systems), and adds more systems and examples. We should probably split this further for proper review once we're happy with the result.

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/2)


hidmic:
@IanTheEngineer I was heading towards a visualization of Drake's `ManipulationStation` example in RViz to show `/tf` broadcasting, but it turns out I can't without upstream changes. SDF files in Drake use relative paths for URIs whereas `rviz` needs a full URLs that `curl` can take OR a `package://` URL (like the one used in #1) to lookup the `ament_index`. 

So I'll stop here, and try with markers instead.

hidmic:
Alright, the `iiwa_manipulator` example is now working. Overall API probably needs a revisit to make sure it's general and that it makes for a good UX.

IanTheEngineer:
Thanks @hidmic! I'll take a pass on testing and come back to comment.

drake_ros_systems/package.xml
```diff
@@ -0,0 +1,25 @@
...
+  <depend>rclcpp</depend>
+  <depend>rosidl_runtime_c</depend>
+  <depend>rosidl_typesupport_cpp</depend>
+  <depend>tf2_ros</depend>
+  <depend>visualization_msgs</depend>
+
+  <test_depend>ament_cmake_gtest</test_depend>
+  <test_depend>ament_lint_auto</test_depend>
+  <test_depend>ament_lint_common</test_depend>
+  <test_depend>test_msgs</test_depend>
```
gbalke, line 20: Where does this come from?

gbalke:
I tested out the flip-flop example and it works great! I put together https://github.com/gbalke/docker-drake-ros2-opengl for testing locally/remotely.

I'm not exactly sure how to get the iiwa_manipulator demo working as it doesn't seem to be the same interface as what @sloretz put together. Could you update the README to describe the startup procedure?

hidmic:
@gbalke README updated!

gbalke:
Hey @hidmic, thanks for the update! I got it working. My only concern with a static demo is that it can be hard to verify things are working correctly as this could just be the starting state of the URDF. It'd be helpful to have a little motion (at the minimum the base rotating sinusoidally between the two bins). This would make it a much quicker visual inspection that things are working as intended. I do see that the topics are echoing correctly though!

drake_ros_systems/README.md
```diff
@@ -0,0 +1,155 @@
...
+ros2 topic pub /S std_msgs/msg/Bool "data: false"
+ros2 topic pub /S std_msgs/msg/Bool "data: true"
+```
+
+```
+ros2 topic pub /R std_msgs/msg/Bool "data: false"
+ros2 topic pub /R std_msgs/msg/Bool "data: true"
+```
+
+## IIWA Manipulator
```
IanTheEngineer, line 120: Not necessary for this PR, but I envision these examples and tutorials to be in separate markdown files, linked to from this main README.

IanTheEngineer (commented):
Checkpoint. This deserves a much more in depth review, and I need to get it running in docker as I am on 18.04 currently.

One high level thought: could the examples be moved to their own package `drake_ros_examples`, which can depend on `drake_ros_systems`? That would reduce the number of `systems` dependencies, and could provide a clearer delineation between "API" and "example".

drake_ros_systems/CMakeLists.txt
```diff
@@ -0,0 +1,124 @@
+cmake_minimum_required(VERSION 3.10)
+project(drake_ros_systems)
+
+# Default to C99
+if(NOT CMAKE_C_STANDARD)
+  set(CMAKE_C_STANDARD 99)
```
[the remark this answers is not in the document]
hidmic: Typical ROS 2 `CMakeLists.txt` prologue, not necessary here. Dropped it.

drake_ros_systems/README.md
hidmic: Done, see `drake_ros_examples` package.

hidmic:
> One high level thought: could the examples be moved to their own package `drake_ros_examples`, which can depend on `drake_ros_systems`? 

Done !

hidmic:
> My only concern with a static demo is that it can be hard to verify things are working correctly as this could just be the starting state of the URDF. It'd be helpful to have a little motion (at the minimum the base rotating sinusoidally between the two bins). This would make it a much quicker visual inspection that things are working as intended.

That's fair. I can work on adding some motion to the arm. @IanTheEngineer thoughts?

hidmic:
@gbalke https://github.com/RobotLocomotion/drake-ros/pull/2/commits/67ed30b1cc865ee72567ab477ec89bcf11bf6040 adds some swing to the base arm joint.

gbalke:
> @gbalke [67ed30b](https://github.com/RobotLocomotion/drake-ros/commit/67ed30b1cc865ee72567ab477ec89bcf11bf6040) adds some swing to the base arm joint.

Verified this works! Thanks :smile:

drake_ros_systems/include/drake_ros_systems/drake_ros.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+namespace drake_ros_systems
+{
+/// PIMPL forward declarations
+class DrakeRosPrivate;
+class Publisher;
+class Subscription;
+
+/// System that abstracts working with ROS
+class DrakeRos final : public DrakeRosInterface
```
IanTheEngineer, line 38: I'm curious as to your thoughts on class hiearchy, why create a high level class like `DrakeRos` to handle publishing, subscribing, clock, and transforms, rather than individual wrappers?

IanTheEngineer (changes requested):
Checkpoint 30/48 thoroughly inspected. Overall, this is a very exciting start to our drake-ros interface. Thanks Miche and & Shane! I would like to see the copyrights updated with the current year at the top of each file, unless you have a reason to keep them dated last year. I will continue to read through and test the rest of the examples later this week.

IanTheEngineer:
@EricCousineau-TRI, if you have any free cycles, could you give the python bindings a look through?

jwnimmer-tri:
@IanTheEngineer I saw you tagged me above.  From my own perspective, I had been interpreting this PR as an illustration to help facilitate discussions about the potential project structure and things we could accomplish.  I think it _is_ helpful to discuss the gross repository structure using a prototype such as this.  For example, the "could the examples be moved to their own package drake_ros_examples" was a perfectly suitable comment to help suss out our plans here.  Having those discussions in the abstract is often difficult, so referring to a working prototype is much more valuable.

However, some of your recent comments upthread are detailed enough to make me wonder if intention here is to merge this PR intact, as a whole?  I don't see how merging 4500+ lines of code in a single PR without CI, without a list of targeted compatibility platforms, without a style guide and linters, without a policy on reviewer responsibilities nor suitably threaded code review tool (such as Reviewable), is going to help move this project forward.  For most of those things, I think that trying to add them after the fact will be frustrating when you already have thousands of lines of code in play to go back and re-review and fix up.

In particular (on the discussion you tagged me on), for things like "let's carefully inspect the copyright of existing code we are copying into here", that's absolutely the kind of there where moving slowly with small groups of files will lead to a much more accurate result than trying to jam it in all at once.

IanTheEngineer:
Thanks for your comment, Jeremy. Along the lines of your initial thoughts, is very much a prototype pull request; one that should represent what is possible. Whether this prototype continues to be iterated upon in the current PR form, or merged into a "rolling development" branch should not change it's experimental status. 

My comments above were a first pass at some of the higher level repository organizational questions you noted, while also incorporating some more nitty-gritty details that I think we should address. Those details were noted as I proceeded so that they don't get lost in this process. I saw my requests for issue creation as a form of note-taking. I don't think PR is near the "API Stable" level we should be accepting as part of Drake, or advertising to folks. I would want Drake reviewers and users feedback first.

That being said, we have to start somewhere, and having the code gathered here in a development branch (to my mind) is better for prototyping and testing than in small one-off repositories or personal forks. I would not mind using some of the CI/Review tools you mentioned, even in this prototype phase.

Then the question becomes, "how do we transition from a functional prototype to a more principled Drake API?". I definitely welcome feedback here. My current thinking is to incorporate the lessons of this prototype into a design document for feedback, with pointers to the prototype for those that would like to try it out. Then we can split (or rewrite) this code into PR's (eg. per class, per ROS package, or per end to end example). I think it makes sense to include some Drake reviewers at this point, also requiring that linters and test coverage make CI green. I am happy to incorporate your thoughts on this plan.

drake_ros_systems/src/ros_publisher_system.cpp
```diff
@@ -0,0 +1,111 @@
...
+  double publish_period)
+: impl_(new RosPublisherSystemPrivate())
+{
+  impl_->serializer_ = std::move(serializer);
+  impl_->pub_ = ros->create_publisher(
+    *impl_->serializer_->get_type_support(), topic_name, qos);
+
+  DeclareAbstractInputPort("message", *(impl_->serializer_->create_default_value()));
+
+  // vvv Mostly copied from LcmPublisherSystem vvv
```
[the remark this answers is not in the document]
mposa: Yes, more than fine with me. Thank you for pushing this effort!

drake_ros_systems/include/drake_ros_systems/rviz_visualizer.hpp
```diff
@@ -0,0 +1,54 @@
...
+/// This system is a subdiagram aggregating a SceneMarkersSystem, a
+/// RosPublisherSystem, and optionally a TfBroadcasterSystem to enable
+/// SceneGraph visualization in RViz. All scene geometries are published
+/// as a visualization_msgs/msg/MarkerArray message to a `/scene_markers`
+/// ROS topic. If `publish_tf` is `true`, all SceneGraph frames are
+/// broadcasted as tf2 transforms.
+///
+/// It exports one input port:
+/// - *graph_query* (abstract): expects a QueryObject from the SceneGraph.
+class RvizVisualizer : public drake::systems::Diagram<double>
```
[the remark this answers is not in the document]
hidmic: > The name RvizVisualizer implies that this method could incorporate more than just MarkerArray's; I could see RobotModel or InteractiveMarker's falling into this category. 

True, and I cannot say I've thought through all the implications. Assuming we can pull that off, I think it's practical to have a drop-in replacement for the `DrakeVisualizer` system. It certainly makes for a smooth migration of existing diagrams.

drake_ros_systems/src/ros_subscriber_system.cpp
```diff
@@ -0,0 +1,130 @@
...
+RosSubscriberSystem::DoCalcNextUpdateTime(
+  const drake::systems::Context<double> & context,
+  drake::systems::CompositeEventCollection<double> * events,
+  double * time) const
+{
+  // Vvv Copied from LcmSubscriberSystem vvv
+
+  // We do not support events other than our own message timing events.
+  LeafSystem<double>::DoCalcNextUpdateTime(context, events, time);
+  DRAKE_THROW_UNLESS(events->HasEvents() == false);
```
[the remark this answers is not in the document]
hidmic: I subjectively think that, since this a Drake system, it makes sense to use Drake APIs. But I'll let @sloretz comment on this.

drake_ros_systems/src/rviz_visualizer.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include <unordered_set>
+
+#include "drake_ros_systems/ros_publisher_system.hpp"
+#include "drake_ros_systems/scene_markers_system.hpp"
+#include "drake_ros_systems/tf_broadcaster_system.hpp"
+
+#include "drake_ros_systems/rviz_visualizer.hpp"
+
+
+namespace drake_ros_systems
```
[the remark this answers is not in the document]
hidmic: Agreed.

drake_ros_systems/src/tf_broadcaster_system.cpp
```diff
@@ -0,0 +1,139 @@
...
+  double publish_period)
+: impl_(new TfBroadcasterSystemPrivate())
+{
+  impl_->tf_broadcaster_ = ros->create_tf_broadcaster();
+
+  impl_->graph_query_port_index =
+    this->DeclareAbstractInputPort(
+    "graph_query", drake::Value<drake::geometry::QueryObject<double>>{}).get_index();
+
+  // vvv Mostly copied from LcmPublisherSystem vvv
```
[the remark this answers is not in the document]
hidmic: I could considering `tf2_ros::TransformBroadcaster` is, for the most part, just a publisher. But I purposefully chose to use the official API. If it's implementation ever changes, this system will remain functional.

drake_ros_systems/src/tf_broadcaster_system.cpp
```diff
@@ -0,0 +1,139 @@
...
+}
+
+drake::systems::EventStatus
+TfBroadcasterSystem::PublishFrames(
+  const drake::systems::Context<double> & context) const
+{
+  const drake::geometry::QueryObject<double> & query_object =
+    get_graph_query_port().Eval<drake::geometry::QueryObject<double>>(context);
+  const drake::geometry::SceneGraphInspector<double> & inspector = query_object.inspector();
+  // TODO(hidmic): publish frame transforms w.r.t. to their parent frame
```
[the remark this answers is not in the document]
hidmic: I couldn't find a way to query both the transform of a frame w.r.t. its parent frame *and* the parent frame name. That's why I was forced to "root" every transform and let `tf` do the heavy lifting.

drake_ros_systems/include/drake_ros_systems/drake_ros.hpp
hidmic: I have to defer to @sloretz here, but I presume this was modeled after the `DrakeLcmInterface` type hierarchy. Arguably, it's not the exact same use case, but it gets us easy dependency injection for testing down the road.

drake_ros_systems/include/drake_ros_systems/ros_publisher_system.hpp
```diff
@@ -0,0 +1,101 @@
...
+    const std::unordered_set<drake::systems::TriggerType> & publish_triggers,
+    double publish_period = 0.0)
+  {
+    // Assume C++ typesupport since this is a C++ template function
+    std::unique_ptr<SerializerInterface> serializer = std::make_unique<Serializer<MessageT>>();
+    return std::make_unique<RosPublisherSystem>(
+      serializer, topic_name, qos, ros_interface, publish_triggers, publish_period);
+  }
+
+  RosPublisherSystem(
```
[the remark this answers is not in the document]
hidmic: Alright, should I defer all documentation to another ticket then?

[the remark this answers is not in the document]
IanTheEngineer: Yes, capturing it in a ticket is preferable so we can task it later.

drake_ros_systems/src/tf_broadcaster_system.cpp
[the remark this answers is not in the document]
IanTheEngineer: Ah, that rationale makes sense. Using the official API is preferable in this case then.

jwnimmer-tri:
Skimming the discussions again, I see I was mostly mistaken about the level of nitty-gritty discussion.  There were a few "nits" at the start that threw me off -- later threads are more architecturally oriented.

I don't know that trying to rework this code back into a design document adds much value.  The code speaks for itself, and is generally clearer than an abstract document would be.

I'm also not worried about API stability; of course things will be quite unstable as we bootstrap and figure out what to do here.

The point I'd mainly advocate for is good hygiene, because that is appropriate for any phase of the development lifecycle.  Something like the `type_conversion` functions are small and standalone, and could be a first bit of code to reach the main branch.  You could use that small PR to establish the review process (including deciding which coding conventions to adopt), and land the code without CI -- it's only a few hundred lines.  Then you have a working codebase that you can build & test locally, and next you can turn on basic CI to cover it.  Then you can add linters and any automated checks on top of that, and start working on the CI pipeline that publishes your API reference documentation.  Somewhere along that line, the tooling will be mature enough to start adding in _real_ code, like some of the message pub/sub system classes, one at a time, with unit tests.

My discomfort most revolves around not being privvy to any discussions about getting software development basics in place, as if were being ignored.  Maybe the plans are so obvious that it just hasn't made it into GitHub discussions, and you've talked about it in meetings.  That's fine too, but it might help to place a few breadcrumbs into GitHub here to plot a course.

drake_ros_systems/src/ros_publisher_system.cpp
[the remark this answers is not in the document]
hidmic: What would be the proper way of doing this (i.e the proper way of conveying these bits of code were re-licensed)? Also, I don't see copyrights in Drake's code base so I presume you use commit metadata to track authorship?

drake_ros_examples/package.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>drake_ros_examples</name>
+  <version>0.1.0</version>
+  <description>Examples using Drake systems that interact with ROS 2.</description>
+  <maintainer email="sloretz@openrobotics.org">Shane Loretz</maintainer>
+  <maintainer email="michel@ekumenlabs.com">Michel Hidalgo</maintainer>
```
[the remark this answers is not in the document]
hidmic: Done in https://github.com/RobotLocomotion/drake-ros/pull/2/commits/6e6a4e9d0f1254c33f894bacecc2ed3d2cd525ec.

drake_ros_systems/src/tf_broadcaster_system.cpp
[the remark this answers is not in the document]
calderpg-tri: We've used the publish-all-frames-in-world approach for several years and it's worked fine so far.

drake_ros_systems/src/tf_broadcaster_system.cpp
```diff
@@ -0,0 +1,139 @@
...
+    transforms.reserve(inspector.num_frames() - 1);
+    geometry_msgs::msg::TransformStamped transform;
+    transform.header.stamp =
+      rclcpp::Time() + rclcpp::Duration::from_seconds(context.get_time());
+    transform.header.frame_id = inspector.GetName(inspector.world_frame_id());
+    for (const drake::geometry::FrameId & frame_id : inspector.all_frame_ids()) {
+      if (frame_id == inspector.world_frame_id()) {
+        continue;
+      }
+      transform.child_frame_id = inspector.GetName(frame_id);
```
calderpg-tri, line 129: What does this produce for frame names of identically-named frames in multiple model instances?

drake_ros_systems/src/tf_broadcaster_system.cpp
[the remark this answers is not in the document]
calderpg-tri: One argument against using the ROS1 TransformBroadcaster was that it was surprisingly inefficient at batching TF update messages and instead sent each transform individually. If that's still the case, I would strongly favor our own publisher-based implementation that batches them together (at least until fixing the upsteam version to batch).

drake_ros_systems/src/tf_broadcaster_system.cpp
hidmic: Good question. I did not test that particular case. Having said that, AFAIK the `SceneGraph` does not seem to have a concept of model. Source systems like the `MultibodyPlant` use prefixes for geometry and frame names e.g. `robot_0::leg_link`. 

Thus,

> identically-named frames in multiple model instances?

shouldn't be a problem.

hidmic: On a related note, I considered substituting `::` with `/` to better align with ROS conventions for frame names, but I figured it'd be putting unnecessary cognitive load on a dev trying to correlate frames in the `SceneGraph` and `/tf`.

drake_ros_systems/src/scene_markers_system.cpp
```diff
@@ -0,0 +1,352 @@
...
+    visualization_msgs::msg::MarkerArray * marker_array)
+  {
+    DRAKE_ASSERT(nullptr != marker_array);
+
+    marker_array_ = marker_array;
+
+    prototype_marker_.header.frame_id =
+      inspector.GetName(inspector.GetFrameId(geometry_id));
+    prototype_marker_.ns = inspector.GetOwningSourceName(geometry_id) +
+      "::" + inspector.GetName(geometry_id);
```
calderpg-tri, line 69: Isn't this going to generate a separate namespace for each marker? That's not very usable if people are trying to visualize a larger system and want to select/deselect chunks of it.

drake_ros_systems/src/tf_broadcaster_system.cpp
calderpg-tri: Multiple model instances with identically-named frames is a very common case in anzu, so we need to make sure this works reasonably. This is part of the reason the anzu MbP+SG->Rviz process uses the MbP names, where model instances are known. (Also note that only some of these items (MbP bodies, MbP frames, SG frames) are guaranteed to have unique names, let alone names at all)

> considered substituting :: with / to better align with ROS conventions for frame names

We've used something like this in anzu
```
std::string GetTfFrameName(const std::string& model_instance_name, const std::string& element_name) {
  const std::regex mbp_separator_regex("::");
  std::string raw_name = model_instance_name + "/" + element_name;
  return std::regex_replace(raw_name, mbp_separator_regex, "/");
}
```
to do the conversion, and I think having ROS-like frame names where `::` always becomes `/` is well worth it to be clearer to users from the ROS side of things.

drake_ros_systems/src/scene_markers_system.cpp
hidmic: It will, yes. I considered using source names as marker namespaces and geometry IDs as marker IDs, but:

a. Geometry IDs are 64 bit integers whereas marker IDs are 32 bit integers. This is relatively minor, as I'd expect things to fall apart way before one can attempt to visualize 2^31 + 1 markers.
b. To render a single, capsule-shaped geometry, three (3) markers are necessary. And with that, the correlation between marker ID and geometry ID is lost.

I guess we could allocate 29 bits for the ID, 2 bits to distinguish capsule geometries' constitutive markers, and throw if bit truncation of the geometry ID would lose information. Then we could afford putting all markers for all geometries from a given source in its namespace. WDYT?

drake_ros_systems/src/tf_broadcaster_system.cpp
hidmic: > This is part of the reason the anzu MbP+SG->Rviz process uses the MbP names, where model instances are known.

> Also note that only some of these items (MbP bodies, MbP frames, SG frames) are guaranteed to have unique names, let alone names at all

I see. Indeed there's [no guarantee](https://github.com/RobotLocomotion/drake/blob/e59b7fc18dbe80b827d07e4a3283a0c87eda7021/geometry/geometry_state.cc#L544-L546) `SceneGraph` frame names are unique... Only frame IDs are unique.

@IanTheEngineer this is a bit of a problem. We cannot

> convert SceneGraph QueryObjects to [...] TF messages

if frame names are not unique. Well, we _could_ generate frame names like "frame_ID" but it is not very practical. We need additional information e.g. by querying the `MultibodyPlant` as @calderpg-tri suggests. I'm curious though, how is the scene described by a `SceneGraph` supposed to be visualized and introspected if the only non-ambiguous identification metadata is a number? I understand frame names become less useful as you scale into the hundredths or thousands of frames, but there must be some semantic metadata somewhere to make sense out of the scene, no?

> I think having ROS-like frame names where :: always becomes / is well worth it to be clearer to users from the ROS side of things.

That's fair. We can do the conversion.

calderpg-tri: See `planning/rviz_geometry_visualizer.{cc, h}` and `planning/shape_description.{cc, h}` in anzu if you haven't already, to see how we've done this. We fall back to something like `<model_instance>/unnamed_frame_id_<frame_id>` in the case of a frame that doesn't have a name, but that's usually only a few frames in a complex system.

drake_ros_systems/src/scene_markers_system.cpp
calderpg-tri: We've made the choice that marker namespaces are useful, but ids within that namespace are entirely opaque and don't have meaning - basically the approach of
```
marker.id = static_cast<int>(marker_array.markers.size() + 1);
marker_array.markers.push_back(marker);
```
In the general "visualize a whole SceneGraph" case, we use namespace = source name, and in other cases we break it down by body/frame names to allow markers corresponding to specific bodies to be selected. See `SceneGraphCollisionChecker::MakeContextRvizRepresentationOfConfigurationCollisions` for the most extreme case of this.

hidmic: That's reasonable approach, but it won't work as-is if the geometries are being added and removed dynamically from the `SceneGraph`. Neither will this patch, though it's easy to adapt it (i.e. to track which geometry IDs have been published already).

hidmic: That's an interesting approach. Have you considered submitting that patch to upstream Rviz?

drake_ros_systems/src/tf_broadcaster_system.cpp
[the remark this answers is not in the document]
hidmic: `tf2`'s TransformBroadcaster does provide an API to send a collection of messages at once, so I'll guess that by batching you mean the ability to aggregate transforms sent one by one in a given time window (which is indeed not supported). If that's the case, I'd much rather make a feature contribution upstream than to re-implement it here.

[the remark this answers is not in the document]
calderpg-tri: Yes, `TransformBroadcaster::sendTransform(const std::vector<geometry_msgs::msg::TransformStamped> & msgtf)` should be good enough for our uses.

Although, I would say in terms of code complexity, `TransformBroadcaster` is literally a publisher with unnecessarily verbose sugar for copying a vector. We should not duplicate code here just because - TF broadcasting is not special, so we should use the `RosPublisherSystem`.

drake_ros_systems/src/tf_broadcaster_system.cpp
IanTheEngineer: In my mind, this is as much a ROS/RViz issue as it is for the `drake-ros` interface. Even in RViz, without special namespace handling, you cannot have multiple instances of robots (having the same names in their frames) robot publishing to `/tf`, and therefore visualized in RViz. It's not immediately obvious to me how we should handle this to make both Drake & `tf` happy. 

I think querying MBP for more information could work for now. We should make note if there is a particular MBP API that would make this process easier, we can propose it to the Drake Developers. From the `tf`/RViz side, we could bring back to life something akin to `tf_prefix`, which was apparently only [deprecated due to a lack of development cycles](http://wiki.ros.org/tf2/Migration#tf_prefix_backwards_compatibility).

drake_ros_systems/src/tf_broadcaster_system.cpp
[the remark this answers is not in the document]
hidmic: Alright. I had to check whether the QoS settings that `tf2_ros` expects were exposed. [They are](https://github.com/ros2/geometry2/blob/ros2/tf2_ros/include/tf2_ros/qos.hpp). We can configure our own publisher with them. We can revisit this if it (ever) breaks.

drake_ros_systems/src/scene_markers_system.cpp
hidmic: https://github.com/ros2/rviz/pull/685 was merged and released into Rolling. Feature's there.

gbalke:
When working on some work for hydroelastic collisions I found out that there are some dimension translation discontinuities between ros markers and drake objects. I'm guessing there's an issue with the translations defined here: https://github.com/RobotLocomotion/drake-ros/blob/hidmic/generic-systems/drake_ros_systems/src/scene_markers_system.cpp


https://user-images.githubusercontent.com/11024792/118735682-c7175e00-b7f5-11eb-8b8f-9c6315450788.mp4

My example is available here: https://github.com/gbalke/drake-ros/tree/generic-systems-extended

drake_ros_systems/src/tf_broadcaster_system.cpp
hidmic: Alright, https://github.com/RobotLocomotion/drake-ros/pull/2/commits/b365190165eee7d24751b273fe432b9d96609b22 and https://github.com/RobotLocomotion/drake-ros/pull/2/commits/1f69048dd13d918cbdaed04471272fba7d786f8d now allows `MultibodyPlant` instances in the scene to be registered with the `RvizVisualizer` to help refine frame IDs and marker namespaces.

drake_ros_systems/src/tf_broadcaster_system.cpp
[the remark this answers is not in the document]
hidmic: See https://github.com/RobotLocomotion/drake-ros/pull/2/commits/5e711c2e9808eeb9cd18098ba09759e4b3b7b014 for this.

hidmic:
> I found out that there are some dimension translation discontinuities between ros markers and drake objects

Wrong marker scales! See https://github.com/RobotLocomotion/drake-ros/pull/2/commits/f01c6c3c6326c193e3fd0f174ba55b4d371f59cd. Thanks @gbalke.

gbalke:
> Wrong marker scales! See f01c6c3. Thanks @gbalke.

Looking great, thanks @hidmic!
image

sloretz:
All content from this PR has been merged as of #39. Closing :tada:

PR #17, opened 2021-08-20
Fix Subscription class in Galactic

### LINKED ISSUE #14: Fix build on ros-galactic-ros-base

hidmic:
Closes #14. Use `virtual` keyword instead of `override` keyword.

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/17)

### DIFF
--- a/drake_ros_core/src/subscription.hpp
+++ b/drake_ros_core/src/subscription.hpp
@@ -62,6 +62,7 @@ protected:
   handle_loaned_message(
     void * loaned_message, const rclcpp::MessageInfo & message_info) override;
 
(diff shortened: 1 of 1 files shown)


hidmic:
This is arguably a hack, but it does the trick and it is less intrusive than conditional compilation (which ROS 2 doesn't provide for).

sloretz (approved):
It works. I would recommend a comment saying it should be changed back to `override` when `galactic` support is no longer needed.

hidmic:
Comment included. Going in !

PR #21, opened 2021-08-25
Add Python bindings for drake_ros_tf2 package

hidmic:
Spin off from #2. Built on top of #20.

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/21)


hidmic (commented):
@aaronchongth it's all yours :) 

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/21#-:-Mvz6FOc14aj6pelBESn:bfi5mw8)* status: all files reviewed, 3 unresolved discussions (waiting on @IanTheEngineer and @sloretz)

___
*[drake_ros_tf2/CMakeLists.txt, line 81 at r8](https://reviewable.io/reviews/robotlocomotion/drake-ros/21#-MvyLNmw7YHWXSmWWjou:-Mvz1zyq9dM-br0eiSe_:b3svzla) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/99fb730a45b9c50c1926576a486eceeb957a1964/drake_ros_tf2/CMakeLists.txt#L81)):*

I think this should use the public function `ament_get_python_install_dir(...)` instead

https://github.com/RobotLocomotion/drake-ros/blob/a0a2932219e174fadd1d68b8b963f97ee4b7b4e0/drake_ros_core/CMakeLists.txt#L80
</blockquote></details>

Absolutely. That'd be more robust.
___
*[drake_ros_tf2/test/test_tf_broadcaster.py, line 100 at r8](https://reviewable.io/reviews/robotlocomotion/drake-ros/21#-MvyOtVB8tG_EWhkU9XN:-Mvz5wdzEuy6wihDOFEd:bl1tavs) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/99fb730a45b9c50c1926576a486eceeb957a1964/drake_ros_tf2/test/test_tf_broadcaster.py#L100)):*

Is this spin_once to handle the callback for the transform listener? If there are any timers internal to rclpy or the executor, it might be possible for the first callback to not be the transform listener. I would recommend a loop here that spins and checks if the transform is ready, maybe with a small timeout or max number of spins.
</blockquote></details>

That's right and yes, it's flaky. Spinning in a loop sounds good.

BTW `rcl*` libraries should have some API to do just that.
___
*[drake_ros_tf2/src/python_bindings/drake_ros_tf2.cc, line 43 at r8](https://reviewable.io/reviews/robotlocomotion/drake-ros/21#-MvyMvRXCNngIavWq7g-:-Mvz3S7746Ux6u0bqifw:brb4uim) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/99fb730a45b9c50c1926576a486eceeb957a1964/drake_ros_tf2/src/python_bindings/drake_ros_tf2.cc#L43)):*

My pybind11 knowledge is a little rusty, but it seems like this may not be the right [return value policy](https://pybind11.readthedocs.io/en/stable/advanced/functions.html#return-value-policies)? It looks like `reference` tells pybind11 not to give the Python object ownership of `obj`, but `obj` is a local variable that will go out of scope when this function returns. I think the default value of `automatic` would fit here.
</blockquote></details>

Looking at latest `pybind11`, using `automatic_reference` (the default) will induce a copy (https://github.com/pybind/pybind11/blob/522c59ceb27e83750c121d5da3d8b67ac446b754/include/pybind11/cast.h#L1025-L1042) and we want this `setattr` calls to mutate `obj` state. IIRC I based this on https://github.com/RobotLocomotion/drake/blob/e59b7fc18dbe80b827d07e4a3283a0c87eda7021/bindings/pydrake/pydrake_pybind.h#L117-L130 to have this plain struct semantics of sorts.

Alternatively, we could replace all of this by extracting each argument from `kwargs` explicitly OR specifying optional keyword arguments for the constructor explicitly. I don't mind either way.

sloretz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/21)* status: all files reviewed, 3 unresolved discussions (waiting on @hidmic, @IanTheEngineer, and @sloretz)

___
*[drake_ros_tf2/src/python_bindings/drake_ros_tf2.cc, line 43 at r8](https://reviewable.io/reviews/robotlocomotion/drake-ros/21#-MvyMvRXCNngIavWq7g-:-MvzP0le3624djkNXEfB:b248ekd) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/99fb730a45b9c50c1926576a486eceeb957a1964/drake_ros_tf2/src/python_bindings/drake_ros_tf2.cc#L43)):*
<details><summary><i>Previously, hidmic (Michel Hidalgo) wrote…</i></summary><blockquote>

Looking at latest `pybind11`, using `automatic_reference` (the default) will induce a copy (https://github.com/pybind/pybind11/blob/522c59ceb27e83750c121d5da3d8b67ac446b754/include/pybind11/cast.h#L1025-L1042) and we want this `setattr` calls to mutate `obj` state. IIRC I based this on https://github.com/RobotLocomotion/drake/blob/e59b7fc18dbe80b827d07e4a3283a0c87eda7021/bindings/pydrake/pydrake_pybind.h#L117-L130 to have this plain struct semantics of sorts.

Alternatively, we could replace all of this by extracting each argument from `kwargs` explicitly OR specifying optional keyword arguments for the constructor explicitly. I don't mind either way.
</blockquote></details>

Ohhhh I thought `pyobj` was the one being returned, but it's `obj` being returned. I understand now.

Since this is the user-facing module, I'd lean towards using `py::kw_only()` and `py::arg(...)` for each keyword argument so that `help(SceneTfBroadcasterParams)`'s would show all the params that can be passed to `__init__()`. I wouldn't block on that though.

sloretz (approved):
:lgtm:

> Reviewed 2 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/21#-:-Mw2-gGS1Qcm1H_JWyxA:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @hidmic and @IanTheEngineer)

aaronchongth (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/21)* status: 24 of 25 files reviewed, 1 unresolved discussion (waiting on @IanTheEngineer and @sloretz)

___
*[drake_ros_tf2/src/python_bindings/drake_ros_tf2.cc, line 43 at r8](https://reviewable.io/reviews/robotlocomotion/drake-ros/21#-MvyMvRXCNngIavWq7g-:-Mw4a9yfD-LOsX2QWKTJ:b7n8vr) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/99fb730a45b9c50c1926576a486eceeb957a1964/drake_ros_tf2/src/python_bindings/drake_ros_tf2.cc#L43)):*

Ohhhh I though `pyobj` was the one being returned, but it's `obj` being returned. I understand now.

Since this is the user-facing module, I'd lean towards using `py::kw_only()` and `py::arg(...)` for each keyword argument so that `help(SceneTfBroadcasterParams)`'s would show all the params that can be passed to `__init__()`. I wouldn't block on that though.
</blockquote></details>

Using `py::kw_only` and `py::arg` in https://github.com/RobotLocomotion/drake-ros/pull/21/commits/d199b4d4636da7e7154096e2027c77b9dcf79868, the `help(...)` function shows the default values now too

PR #23, opened 2021-08-26
Add Drake C++ systems to drake_ros_viz package.

hidmic:
Spin off form #2. Built on top of #22.

To do:
- [x] Add tests for marker IDs when there are multiple markers using the same marker namespace. The IDs should be incremented for each instance.

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/23)

PR #25, opened 2021-11-02
Add Hydroelastic Collision System

gbalke:
Builds off of #23. Adds hydroelastic collision visualization. The textures are not coming through on current rolling as 8.7.0 was released right before I added the textures. I can bump rviz maintainers to make a release soon if we want this in ASAP.

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/25)


IanTheEngineer:
@gbalke can you retarget this PR at `develop` please?

gbalke:
To note, there is a big diff from bringing in [lodepng](https://github.com/lvandeve/lodepng) which does the raw bitmap to image conversion before transmission to ROS. It's a pretty big file but it is a copy/paste. I don't know how you handle 3rd party libraries so that might be a point of discussion.

gbalke:
To run the whole setup, I'm using my drake-ros docker image https://github.com/gbalke/docker-drake-ros2-opengl

After following setup instructions, make a ros workspace with drake-ros in it inside of the `dev` folder. Build this inside of the container bash environment. Next, run the following to bring up the visualizers:
```
# Run this from within the docker container. In this case, my colcon ws is in dev/ros2/
rviz2 rviz2 -d dev/ros2/src/drake-ros/drake_ros_viz/examples/hydroelastic_collisions/hydroelastic_collisions.rviz

# Run this from the drake-ros-docker project folder
./drake_viz.sh
```

Finally, source the built workspace from within the docker bash environment and run:
```
ros2 run drake_ros_viz hydroelastic_collisions
```

You should now see the following:
image

Texture color will be added with the next rviz release. This is what it looks like when compiling rviz from source (`ros2` branch):
image

gbalke:
@IanTheEngineer I'm updating my docker-drake-ros2-opengl repo to docker-ros2-opengl. I'm going to instead maintain branches with add-ins so the branch https://github.com/gbalke/docker-ros2-opengl/tree/ros2-drake will have the ros2-drake integration and the ros2 branch will just have ros2 (rolling). This way the work can be re-used for other integration projects.

PR #32, opened 2022-01-12
[POC] Add drake_ros_introspection package

hidmic:
This is a proof of concept of an introspection mechanism for Drake systems by exposing their data flows via ROS 2 interfaces. To do this, it constructs a `drake::systems::Simulator` monitor. At the moment, only input and output ports can be monitored using ROS 2 topics, but it can be readily extended to monitor system state.

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/32)

### DIFF
--- a/drake_ros_introspection/include/drake_ros_introspection/predicates/declared_by.h
+++ b/drake_ros_introspection/include/drake_ros_introspection/predicates/declared_by.h
@@ -1,0 +1,104 @@
+// Copyright 2022 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <algorithm>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include <drake/systems/framework/system_base.h>
+
+namespace drake_ros_introspection {
+namespace predicates {
+
+namespace internal {
+
+template <typename BaseT, typename T>
+bool isinstance(const T& value) {
+  return (dynamic_cast<const BaseT*>(&value) != nullptr);
+}
+
+template <typename BaseT, typename NextBaseT, typename... BasesT, typename T>
+bool isinstance(const T& value) {
+  return isinstance<BaseT>(value) || isinstance<NextBaseT, BasesT...>(value);
+}
+
+template <template <typename> class Base, template <typename> class... Bases>
+class IsTemplateInstanceOf {
+ public:
+  template <template <typename> class Class, typename... ArgsT>
+  bool operator()(const Class<ArgsT...>& instance) const {
+    return isinstance<Base<ArgsT...>, Bases<ArgsT...>...>(instance);
+  }
+};
+
+template <typename T>
+class IsOneOf {
+ public:
+  explicit IsOneOf(std::initializer_list<T*> instances)
+      : instances_(instances) {}
+
+  bool operator()(const T& instance) const {
+    auto it = std::find(instances_.begin(), instances_.end(), &instance);
+    return it != instances_.end();
+  }
+
+ private:
+  const std::vector<T*> instances_;
+};
+
+class SystemNameMatches {
+ public:
+  explicit SystemNameMatches(std::initializer_list<std::string> names)
+      : names_(names) {}
+
+  bool operator()(const drake::systems::SystemBase& instance) const {
+    auto it = std::find(names_.begin(), names_.end(), instance.get_name());
+    if (it == names_.end()) {
+      it =
+          std::find(names_.begin(), names_.end(), instance.GetSystemPathname());
+    }
+    return it != names_.end();
+  }
+
+ private:
+  std::vector<std::string> names_;
+};
+
+}  // namespace internal
+
+template <template <typename> class System,
+          template <typename> class... Systems>
+internal::IsTemplateInstanceOf<System, Systems...> DeclaredBy() {
+  return internal::IsTemplateInstanceOf<System, Systems...>{};
+}
+
+template <typename SystemT, typename... SystemsT>
+internal::IsOneOf<const drake::systems::SystemBase> DeclaredBy(
+    const SystemT* system, const SystemsT*... systems) {
+  return internal::IsOneOf<const drake::systems::SystemBase>{system,
+                                                             systems...};
+}
+
+template <typename NameT, typename... NamesT>
+internal::SystemNameMatches DeclaredBy(const NameT& name,
+                                       const NamesT&... names) {
+  return internal::SystemNameMatches{name, names...};
+}
+
+}  // namespace predicates
+
+}  // namespace drake_ros_introspection
--- a/drake_ros_introspection/include/drake_ros_introspection/probes/port_probe.h
+++ b/drake_ros_introspection/include/drake_ros_introspection/probes/port_probe.h
@@ -1,0 +1,90 @@
+// Copyright 2022 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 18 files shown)


hidmic:
@IanTheEngineer @sloretz I included an `introspection_demo` executable (for demonstration and manual testing), but it is still quite simple and doesn't fully exercise the library. I'll improve it next. In the meantime, PTAL!

IanTheEngineer (commented):
For future features: Introspect State, Graph IO's, and potentially plot.

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/32#-:-MtPA_fE0CuNl6UQZXz3:biozv35)* status: 0 of 18 files reviewed, all discussions resolved (waiting on @sloretz)

drake_ros_introspection/include/drake_ros_introspection/predicates/declared_by.h
```diff
@@ -0,0 +1,104 @@
...
+
+#include <drake/systems/framework/system_base.h>
+
+namespace drake_ros_introspection {
+namespace predicates {
+
+namespace internal {
+
+template <typename BaseT, typename T>
+bool isinstance(const T& value) {
```
sloretz, line 30: `constexpr bool` since this is a compime check?

drake_ros_introspection/include/drake_ros_introspection/predicates/declared_by.h
```diff
@@ -0,0 +1,104 @@
...
+  return (dynamic_cast<const BaseT*>(&value) != nullptr);
```
sloretz, line 31: Is this the same as `return std::is_base_of<BaseT, T>::value`? If not, I'm surprised there isn't something in `type_traits` for it.

drake_ros_introspection/include/drake_ros_introspection/predicates/declared_by.h
```diff
@@ -0,0 +1,104 @@
...
+    auto it = std::find(names_.begin(), names_.end(), instance.get_name());
+    if (it == names_.end()) {
+      it =
+          std::find(names_.begin(), names_.end(), instance.GetSystemPathname());
+    }
+    return it != names_.end();
+  }
+
+ private:
+  std::vector<std::string> names_;
```
sloretz, line 78: Are there expected to be a lot of system names? If so, maybe `unordered_set` could give faster lookup.

drake_ros_introspection/include/drake_ros_introspection/probes/port_probe.h
```diff
@@ -0,0 +1,90 @@
...
+      const drake::systems::Context<ScalarT>& context) override {
+    DRAKE_THROW_UNLESS(publisher_ != nullptr);
+    if (publisher_->get_subscription_count() > 0) {
+      const PortT& port = this->get_port();
+      const drake::systems::Context<ScalarT>& subcontext =
+          port.get_system().GetMyContextFromRoot(context);
+      using convert = utilities::convert<ValueT, ValueConvention, MessageT>;
+      publisher_->publish(
+          convert::to_message(port.template Eval<ValueT>(subcontext)));
+      return drake::systems::EventStatus::Succeeded();
```
sloretz, line 67: If the `Probe` isn't changing the state in the `Context`, mabye `DidNothing` is the appropriate return? The [docs say](https://drake.mit.edu/doxygen_cxx/classdrake_1_1systems_1_1_event_status.html#a11717da96dc4e4a5f79371318448f82b):

> kDidNothing | Successful, but nothing happened; no state update needed.

which makes me think returning `Succeeded` will cause `Simulator` to do extra work.

sloretz (commented):
I'm still reviewing; I don't understand how it all works yet. Here are a few comments in the meantime

drake_ros_introspection/include/drake_ros_introspection/probes/port_probe.h
hidmic: Ah, good point. [It won't stop stepping](https://github.com/RobotLocomotion/drake/blob/a43f89bc58886452630eccfc15b4bbd5cc79d23f/systems/analysis/simulator.cc#L306) for as long as the monitor reports success (or time runs out).

drake_ros_introspection/include/drake_ros_introspection/predicates/declared_by.h
hidmic: I don't expect many, no. Predicate API itself is based on initializer lists and variadic functions i.e. `DeclaredBy("some_system", "another_system")`, so it already makes it somewhat impractical to pass too many. As I think of it, if a probing rule applies to many systems, then it's quite likely that you want to match by system type instead.

drake_ros_introspection/include/drake_ros_introspection/predicates/declared_by.h
hidmic: Not a compile time check. This function is used to detect whether a `SystemBase` type instance is actually an (upcast) instance of some specific `LeafSystem`, and that can only be done in runtime.

drake_ros_introspection/include/drake_ros_introspection/simulator_monitor.h
```diff
@@ -0,0 +1,55 @@
...
+template <typename T>
+class SimulatorMonitor {
+ public:
+  explicit SimulatorMonitor(
+      std::vector<std::unique_ptr<probes::ProbeInterface<T>>> probes)
+      : probes_(std::move(probes)) {}
+
+  void Configure(const std::shared_ptr<rclcpp::Node>& node) {
+    for (auto& probe : probes_) {
+      probe->Configure(node);
```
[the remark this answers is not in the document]
hidmic: I was a bit on the fence with this, and chose simple over general (it's `drake_ros_introspection` after all, there will be ROS stuff involved). As it stands, it's easy on the user and the API doesn't require the node to be around when probing rules are defined. Yeah, we could push it to `Probe` construction, upon building or indirectly through rules. The latter is the most general but also requires the most plumbing. 

As long as UX doesn't degrade, I'm open to ideas :)

drake_ros_introspection/include/drake_ros_introspection/simulator_monitor.h
```diff
@@ -0,0 +1,55 @@
...
+    }
+  }
+
+  operator std::function<
+      drake::systems::EventStatus(const drake::systems::Context<T>&)>() {
+    return [this](const drake::systems::Context<T>& context) {
+      auto status = drake::systems::EventStatus::DidNothing();
+      for (auto& probe : probes_) {
+        status.KeepMoreSevere(probe->Probe(context));
```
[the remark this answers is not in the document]
hidmic: > If someone has written a pure-Drake "monitor" then it cannot be used without being wrapped in a Probe subclass.

That's true if the user wants to include it within the `SimulatorMonitor` (and delegate event status aggregation), but nothing prevents them from having e.g. a lambda that invokes both the `SimulatorMonitor` and their own functional.

> I think it would be a little simpler as a pure-drake utility for combining multiple `Simulator` "monitor"s

Yeah, I see the value in a utility that helps combine multiple monitor functions.

> Instead of having a list of probes, it would have a list of `std::function< EventStatus(const Context< T > &)>`. Then pure-Drake "monitors" whose job was only to determine when to terminate the simulation could be used as well as `Probe` subsclasses that inspect the system.

Something has to own `Probe` instances though. Thinking out loud, we could go fully functional, sorting out `std::function` copyability requirement via `std::shared_ptr` instances. It was an arbitrary choice to do unique ownership and dynamic polymorphism over shared ownership and type erasure (though I will say that I tend to prefer consistent type systems).

drake_ros_introspection/include/drake_ros_introspection/simulator_monitor_builder.h
```diff
@@ -0,0 +1,120 @@
...
+  class Worker : public drake::systems::SystemVisitor<T> {
+   public:
+    explicit Worker(const Blueprint& blueprint) : blueprint_(blueprint) {}
+
+    void VisitSystem(const drake::systems::System<T>& system) override {
+      BuildInputPortProbes(system);
+      BuildOutputPortProbes(system);
+    }
+
+    void VisitDiagram(const drake::systems::Diagram<T>& diagram) override {
```
sloretz, line 76: VisitSystem + VisitDiagram and therefor BuildInputPortProbes + BuildOutputPortProbes never seem to be used. Are they essential?

hidmic: Those are called (indirectly) from `SimulatorMonitor::Build`. This inner `Worker` class implements the [`SystemVisitor`](https://github.com/RobotLocomotion/drake/blob/master/systems/framework/system_visitor.h#L14) pattern to recursively crawl the given diagram.

EricCousineau-TRI:
Per meeting today, whoever continues this work should look at Anzu's LCM <-> Drake System introspection capabilities (avoid accidental duplication of work / discovery).

EricCousineau-TRI:
Closing in lieu of updated #93, and tracking in #228

PR #39, opened 2022-02-05
Add drake_ros_examples package

aaronchongth:
Spin off from #2. Built on top of #24.

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/39)

PR #46, opened 2022-03-04
Add multirobot example

gbiggs:
This PR adds a multirobot example using Drake-ROS. It creates a two-dimensional array of Kuka LBR iiwa robots, which can then be visualised using rviz.

See the README.md file for how to run the example.

Questions that need answering:

- Should the `iiwa_description` directory be made its own package? (Or, is it already a package somewhere that we can use?)
- Should we add a C++ version of the sample as well?
- Is "the robots all fall down due to gravity" a sufficient example, or should the robots do something?

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/46)


EricCousineau-TRI (changes requested):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/46)* status: 0 of 62 files reviewed, 2 unresolved discussions (waiting on @EricCousineau-TRI and @gbiggs)

*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake-ros/46#-MxVhe0jBYgiM8VtNbY_:-MxVhe0kDvsIArrqRSsM:b-fw6m5j) (no related file):*
> Should the iiwa_description directory be made its own package? (Or, is it already a package somewhere that we can use?)

We should use Drake's version, and not re-vendor it here.

And I believe our version does not use `/robot_description`, so we need not expose it (yet) on the package path.

---
*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake-ros/46#-MxVi9CP9eLHdg9HBCDJ:-MxVi9CP9eLHdg9HBCDK:b7ayrn7) (no related file):*
> Should we add a C++ version of the sample as well?

Nah, I think Python is sufficient!

---
*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake-ros/46#-MxViEOB72klirLrckqB:-MxViEOB72klirLrckqC:b-3n2xg2) (no related file):*
> Is "the robots all fall down due to gravity" a sufficient example, or should the robots do something?

I think it's fine as long as it's clearly indicated somewhere (README) that this is the expected behavior, e.g.
```
This will simulate the array of IIWA robots without control. The expected behavior is for
them to "flop" about as simulation time progresses.
```

---

gbiggs:
@IanTheEngineer Do you have a suggestion on how to find the location of the `iiwa_description` included with Drake? The virtual environment directory could be anywhere.

My first thought is to configure a Python file using CMake so that it contains the path as found by CMake's `find_package(drake)`, but the downside of that is that it is fixed. If the location of the virtual environment containing drake changes, then it will brake.

EricCousineau-TRI:
Depends on #39

EricCousineau-TRI:
From Anzu:

```cc
PackageMap MakeDefaultAnzuPackageMap() {
  PackageMap package_map;

  // Add ability to read `package://anzu` URIs.
  const anzu::filesystem::path packagefile_abspath =
      FindAnzuResourceOrThrow("package.xml");
  package_map.Add("anzu", packagefile_abspath.parent_path());

  // Add packages from drake.
  const anzu::filesystem::path abspath_under_drake_root =
      drake::FindResourceOrThrow("drake/.drake-find_resource-sentinel");
  package_map.PopulateFromFolder(abspath_under_drake_root.parent_path());

  return package_map;
}
```

jwnimmer-tri:
>  // Add packages from drake.

That stanza is no longer relevant (it used to be relevant last year).  The `drake::multibody::PackageMap` by default has the (sole) Drake package already added.  There is no need to add any additional packages from Drake.

EricCousineau-TRI:
ah, kewl. will scrub that from Anzu's code then - thanks!

gbiggs:
Thanks for the tip, @EricCousineau-TRI . I've got it working.

However, somewhere in my rebasing I messed up, somehow, and now this PR wants to remove all of the `develop` branch's content. I've given up trying to clean it up and just made a new branch and PR over at #85.

PR #57, opened 2022-04-06
Make TF frame name for world frame always be world

sloretz:
This PR makes the TF frame for the world frame always be `world`. Currently the frame ends up being `WorldModelInstance/WorldBody/10`. It's not obvious to me if the ` drake::geometry::FrameId` at the end is stable. If not, it would be hard to make an RViz config.

It must have evaluated to `world` at some point in the past because the example PR uses a fixed frame of `world` already in its RViz config.

https://github.com/RobotLocomotion/drake-ros/blob/28b4920e67715eab25fe425601e8fe4529e28a69/drake_ros_examples/examples/iiwa_manipulator/iiwa_manipulator.rviz#L91

@IanTheEngineer & @EricCousineau-TRI thoughts?

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/57)

### DIFF
--- a/drake_ros_tf2/src/utilities/name_conventions.cc
+++ b/drake_ros_tf2/src/utilities/name_conventions.cc
@@ -27,6 +27,11 @@ std::string GetTfFrameName(
     const std::unordered_set<const drake::multibody::MultibodyPlant<double>*>&
         plants,
     const drake::geometry::FrameId& frame_id) {
+  // Special case: world frame is always world
+  if (frame_id == inspector.world_frame_id()) {
+    return "world";
+  }
+
   std::stringstream ss;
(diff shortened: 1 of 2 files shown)


EricCousineau-TRI (approved):
:lgtm_strong: 

I strongly support having the name be correct w.r.t. the existing ecosystem!

> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/57#-:-N-LA7Qg63DhVUTy9lCt:b1dda9l)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @IanTheEngineer)

PR #59, opened 2022-04-08
try to capture failure w/ amdgpu

EricCousineau-TRI:
At presently, `bazel build @ros2//:rviz_rviz` still works?

FYI @sloretz @IanTheEngineer @calderpg-tri

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/59)

PR #61, opened 2022-04-08
Determine system library dirs by asking ld and ldconfig 

### LINKED ISSUE #60: `bazel build @ros2//:rviz_rviz` is misconfigured w/ amdgpu setup?
@calderpg-tri ran into this:
```sh
$ bazel build @ros2//:rviz2_rviz2
ERROR: {bazel_cache}/external/ros2/BUILD.bazel:551:11: @ros2//:rviz_ogre_vendor_cc: invalid label '/opt/amdgpu/lib/x86_64-linux-gnu/libGL.so.1' in element 24 of attribute 'srcs' in 'cc_library' rule: invalid target name '/opt/amdgpu/lib/x86_64-linux-gnu/libGL.so.1': target names may not start with '/'
ERROR: {bazel_cache}/external/ros2/BUILD.bazel:551:11: @ros2//:rviz_ogre_vendor_cc: invalid label '/opt/amdgpu/lib/x86_64-linux-gnu/libGL.so.1' in element 25 of attribute 'data' in 'cc_library' rule: invalid target name '/opt/amdgpu/lib/x86_64-linux-gnu/libGL.so.1': target names may not start with '/'
ERROR: error loading package '@ros2//': Package '' contains errors
...
```

for more info, see #59

sloretz:
Fixes #60

~~I don't understand the problem completely, so I'm not 100% sure this is the right fix.~~

The amdgpu driver adds a file to `/etc/ld.so.conf.d/` to specify a directory in `/opt` that should be searched by the dynamic loader when resolving a shared library's dynamic dependencies (stuff output when doing `ldd libsomelibrary.so`).~~

The CMake scraping code looks at libraries that need to be linked with a given target. Sometimes it knows it needs to link with a system shared library, and somehow it determines that library can be found in one of the folders `/etc/ld.so.conf.d`. This gets reported to code populating the templates for `BUILD.bazel`. That code tries to map the found library to a path in the sandbox, but it can't, so it leaves it as an absolute path to the library. The data is going to be populated into `srcs` and `data` which are bazel labels, and labels can't start with `/`, so the whole `BUILD.bazel` is invalid.

**Edit** There is existing code to filter system libraries, but it uses hard coded directories. This PR uses `ld` and `ldconfig` to determine what the system directories are instead.

~~This PR resolves it by making the template code filter libraries that are still absolute paths after trying to map them into the sandbox.~~

~~Parts I'm not sure about:~~
* ~~Do the libraries in `/etc/ld.so.conf.d` need to be added to the sandbox, or is `bazel` going to do that automatically?~~
* ~~How did CMake learn that a library needing to be linked against is in a folder specified in a file in `/etc/ld.so.conf.d`?~~
* ~~Where is the `sandbox()` function getting it's data from?~~

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/61)

### DIFF
--- a/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py
+++ b/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py
@@ -3,6 +3,7 @@ Utility functions to aid compilation and linkage configuration scraping by
 resorting to (Linux) system-wide conventions and tooling.
 """
 
+from functools import lru_cache
 import os
 import re
 import subprocess
@@ -25,24 +26,52 @@ def is_system_include(include_path):
         for path in DEFAULT_INCLUDE_DIRECTORIES)
 
 
-# Standard library files' search paths for linkers in Linux systems.
-# Useful to detect system libraries in package exported configuration.
-DEFAULT_LINK_DIRECTORIES = ['/lib', '/usr/lib', '/usr/local/lib']
+@lru_cache(maxsize=1)
+def system_link_dirs():
+    """Get standard search paths for the linker."""
+    link_dirs = set(os.environ.get('LIBRARY_PATH', '').split(':'))
+    output = subprocess.run(
+        ['ld', '--verbose'],
+        check=True,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.DEVNULL,
+        encoding='utf8'
+    ).stdout.strip()
+
+    for directory in re.findall(r'SEARCH_DIR\("=([^=]+)"\)', output):
+        link_dirs.add(directory)
+    # Filter emtpy strings
+    return tuple([d for d in link_dirs if d])
+
+
+@lru_cache(maxsize=1)
+def system_shared_lib_dirs():
+    """Get standard search paths for the dynamic runtime loader."""
+    lib_dirs = set(os.environ.get('LD_LIBRARY_PATH', '').split(':'))
+    output = subprocess.run(
+        ['ldconfig', '--verbose'],
+        check=False,  # TODO(sloretz) why does this proces have a non-zero exit
+                      # code on my system?
+        stdout=subprocess.PIPE,
+        stderr=subprocess.DEVNULL,
+        encoding='utf8'
+    ).stdout.strip()
+
+    for directory in re.findall(r'([^:\t\n]+):', output):
+        lib_dirs.add(directory)
+    # Filter emtpy strings
+    return tuple([d for d in lib_dirs if d])
 
 
 def is_system_library(library_path):
     """
     Checks whether `library_path` is in a system library directory
-    i.e. known to linkers.
+    i.e. known to the linker or dynamic loader
     """
     library_path = os.path.realpath(library_path)
     return any(
         library_path.startswith(path)
-        for path in DEFAULT_LINK_DIRECTORIES)
-
-
-LD_LIBRARY_PATHS = [
-    path for path in os.environ.get('LD_LIBRARY_PATH', '').split(':') if path]
+        for path in system_link_dirs() + system_shared_lib_dirs())
 
 
 def find_library_path(library_name, link_directories=None, link_flags=None):
@@ -66,9 +95,9 @@ def find_library_path(library_name, link_directories=None, link_flags=None):
             cmd.extend(['-L', path])
     if link_flags:
         cmd.extend(link_flags)
-    for path in LD_LIBRARY_PATHS:
+    for path in system_shared_lib_dirs():
         cmd.extend(['-L', path])
-    for path in DEFAULT_LINK_DIRECTORIES:
+    for path in system_link_dirs():
         cmd.extend(['-L', path])
     cmd.extend(['-o', os.devnull, '-l' + library_name])
     try:
--- a/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py
+++ b/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py
@@ -29,7 +29,7 @@ def is_system_include(include_path):
 @lru_cache(maxsize=1)
 def system_link_dirs():
     """Get standard search paths for the linker."""
-    link_dirs = set(os.environ.get('LIBRARY_PATH', '').split(':'))
+    link_dirs = set()
     output = subprocess.run(
         ['ld', '--verbose'],
         check=True,
@@ -47,9 +47,9 @@ def system_link_dirs():
 @lru_cache(maxsize=1)
 def system_shared_lib_dirs():
     """Get standard search paths for the dynamic runtime loader."""
-    lib_dirs = set(os.environ.get('LD_LIBRARY_PATH', '').split(':'))
+    lib_dirs = set()
     output = subprocess.run(
-        ['ldconfig', '--verbose'],
+        ['ldconfig', '-XN', '--verbose'],
         check=False,  # TODO(sloretz) why does this proces have a non-zero exit
                       # code on my system?
         stdout=subprocess.PIPE,
@@ -95,7 +95,7 @@ def find_library_path(library_name, link_directories=None, link_flags=None):
             cmd.extend(['-L', path])
     if link_flags:
         cmd.extend(link_flags)
-    for path in system_shared_lib_dirs():
+    for path in tuple(set(os.environ.get('LIBRARY_PATH', '').split(':'))):
         cmd.extend(['-L', path])
     for path in system_link_dirs():
         cmd.extend(['-L', path])
--- a/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py
+++ b/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py
@@ -88,19 +88,18 @@ def find_library_path(library_name, link_directories=None, link_flags=None):
     # Adapted from ctypes.util.find_library_path() implementation.
     pattern = r'/(?:[^\(\)\s]*/)*lib{}\.[^\(\)\s]*'.format(library_name)
 
-    cmd = ['ld', '-t']
+    paths = []
     if link_directories:
-        for path in link_directories:
-            cmd.extend(['-L', path])
+        paths.extend(link_directories)
+    paths.extend(set(os.environ.get('LIBRARY_PATH', '').split(':')))
+    paths.extend(set(os.environ.get('LD_LIBRARY_PATH', '').split(':')))
+    paths.extend(system_link_dirs())
+    paths.extend(system_shared_lib_dirs())
+
+    cmd = ['ld', '-t']
     if link_flags:
         cmd.extend(link_flags)
-    for path in set(os.environ.get('LIBRARY_PATH', '').split(':')):
-        cmd.extend(['-L', path])
-    for path in set(os.environ.get('LD_LIBRARY_PATH', '').split(':')):
-        cmd.extend(['-L', path])
-    for path in system_link_dirs():
-        cmd.extend(['-L', path])
-    for path in system_shared_lib_dirs():
+    for path in paths:
         cmd.extend(['-L', path])
     cmd.extend(['-o', os.devnull, '-l' + library_name])
     try:
--- a/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py
+++ b/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py
@@ -40,7 +40,7 @@ def system_link_dirs():
 
     for directory in re.findall(r'SEARCH_DIR\("=([^=]+)"\)', output):
         link_dirs.add(directory)
-    # Filter emtpy strings
+    # Filter empty strings
     return tuple([d for d in link_dirs if d])
 
 
@@ -58,7 +58,7 @@ def system_shared_lib_dirs():
 
     for directory in re.findall(r'([^:\t\n]+):', output):
         lib_dirs.add(directory)
-    # Filter emtpy strings
+    # Filter empty strings
     return tuple([d for d in lib_dirs if d])
 
 
--- a/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py
+++ b/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py
@@ -50,8 +50,7 @@ def system_shared_lib_dirs():
     lib_dirs = set()
     output = subprocess.run(
         ['ldconfig', '-XN', '--verbose'],
-        check=False,  # TODO(sloretz) why does this proces have a non-zero exit
-                      # code on my system?
+        check=True,
         stdout=subprocess.PIPE,
         stderr=subprocess.DEVNULL,
         encoding='utf8'
--- a/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py
+++ b/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py
@@ -94,7 +94,7 @@ def find_library_path(library_name, link_directories=None, link_flags=None):
             cmd.extend(['-L', path])
     if link_flags:
         cmd.extend(link_flags)
-    for path in tuple(set(os.environ.get('LIBRARY_PATH', '').split(':'))):
+    for path in set(os.environ.get('LIBRARY_PATH', '').split(':')):
         cmd.extend(['-L', path])
     for path in system_link_dirs():
         cmd.extend(['-L', path])
--- a/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py
+++ b/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py
@@ -96,7 +96,11 @@ def find_library_path(library_name, link_directories=None, link_flags=None):
         cmd.extend(link_flags)
     for path in set(os.environ.get('LIBRARY_PATH', '').split(':')):
         cmd.extend(['-L', path])
+    for path in set(os.environ.get('LD_LIBRARY_PATH', '').split(':')):
+        cmd.extend(['-L', path])
     for path in system_link_dirs():
+        cmd.extend(['-L', path])
+    for path in system_shared_lib_dirs():
         cmd.extend(['-L', path])
     cmd.extend(['-o', os.devnull, '-l' + library_name])
     try:
--- a/bazel_ros2_rules/ros2/ros2bzl/templates.py
+++ b/bazel_ros2_rules/ros2/ros2bzl/templates.py
@@ -64,7 +64,13 @@ def configure_package_cc_library(
     name, metadata, properties, dependencies, sandbox
 ):
     target_name = cc_name(name, metadata)
-    libraries = [sandbox(library) for library in properties['link_libraries']]
+    libraries = [
+        sandbox(library)
+        for library in properties['link_libraries']
+        # Assume only libraries that map to a relative path inside the
+        # sandbox need to be mentioned in BUILD.bazel
+        if not os.path.isabs(sandbox(library))
+    ]
     include_directories = [
         sandbox(include) for include in properties['include_directories']]
     local_includes = [
@@ -188,7 +194,11 @@ def configure_package_py_library(
             config.update({
                 'cc_name': c_name("_" + target_name, metadata),
                 'cc_libs': [
-                    sandbox(lib) for lib in properties['cc_libraries']],
+                    sandbox(lib) for lib in properties['cc_libraries']
+                    # Assume only libraries that  map to a relative path inside
+                    # the sandbox need to be mentioned in BUILD.bazel
+                    if not os.path.isabs(sandbox(lib))
+                ],
                 'cc_deps': cc_deps
             })
             data.append(c_label("_" + target_name, metadata))
--- a/bazel_ros2_rules/ros2/ros2bzl/templates.py
+++ b/bazel_ros2_rules/ros2/ros2bzl/templates.py
@@ -64,13 +64,7 @@ def configure_package_cc_library(
     name, metadata, properties, dependencies, sandbox
 ):
     target_name = cc_name(name, metadata)
-    libraries = [
-        sandbox(library)
-        for library in properties['link_libraries']
-        # Assume only libraries that map to a relative path inside the
-        # sandbox need to be mentioned in BUILD.bazel
-        if not os.path.isabs(sandbox(library))
-    ]
+    libraries = [sandbox(library) for library in properties['link_libraries']]
     include_directories = [
         sandbox(include) for include in properties['include_directories']]
     local_includes = [
@@ -194,11 +188,7 @@ def configure_package_py_library(
             config.update({
                 'cc_name': c_name("_" + target_name, metadata),
                 'cc_libs': [
-                    sandbox(lib) for lib in properties['cc_libraries']
-                    # Assume only libraries that  map to a relative path inside
-                    # the sandbox need to be mentioned in BUILD.bazel
-                    if not os.path.isabs(sandbox(lib))
-                ],
+                    sandbox(lib) for lib in properties['cc_libraries']],
                 'cc_deps': cc_deps
             })
             data.append(c_label("_" + target_name, metadata))


EricCousineau-TRI (changes requested):
> Do the libraries in /etc/ld.so.conf.d need to be added to the sandbox, or is bazel going to do that automatically?

I believe certain directories will not get "mapped out" when running under a nominal Bazel sandbox. For example, I *think* things like `/usr/lib`, `/lib`, etc. will be present under `bazel test`.
(This comes at the cost of less "hermeticity", but I think it should be fine for this case for now. Not everything is encapsulated by the Bazel sandbox, I think).

To confirm, though, we should at least test. I would suggest adding a `bazel test` to repro branch to test for existence of that library under `bazel test`. (added discussion point below)

> How did CMake learn that a library needing to be linked against is in a folder specified in a file in /etc/ld.so.conf.d?
> Where is the sandbox() function getting it's data from?

Probably need to look at the code to posit this? I can help if you'd like, but my attempt to grok would be similar to ours.
Re: CMake, it may be that Miche analyzied something like `ldd`, or the linker flags.
Re: `sandbox()`, I'm sure this is self-contained in this repo.

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-:-N-AOdJ240gXzAPo51mr:byukbg6)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @sloretz)

*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-AOULtCJkJWGnDnTk7:-N-AOULtCJkJWGnDnTk8:b-d3mrvk) (no related file):*
nit At present, I believe only you + Calder can repro this.

Can I ask you to post a draft PR, similar to #59, that shows the repro? I tried on my side w/ your suggestion, but couldn't capture it.

---
*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-AOxJw9jkhBgjcZNGt:-N-AOxJw9jkhBgjcZNGu:b-bohhwy) (no related file):*
Per above, we should have a reproducible confirmation that we can read libraries listed for `/etc/ld.so.conf.d`.
(I believe they should be available, so hopefully no extra work is needed.)

---

sloretz (commented):
> Probably need to look at the code to posit this? I can help if you'd like, but my attempt to grok would be similar to ours.
Re: CMake, it may be that Miche analyzied something like ldd, or the linker flags.

Sure enough it was coming from [using `ldd` to determine each shared libraries dependencies](https://github.com/RobotLocomotion/drake-ros/blob/8ad9e583d48c3507f2346c6b8b6c87491fdf0642/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py#L107). There is logic to filter out system libraries, but [it uses hard-coded paths](https://github.com/RobotLocomotion/drake-ros/blob/8ad9e583d48c3507f2346c6b8b6c87491fdf0642/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py#L30). I updated the PR to get the system directories from `ld` and `ldconfig` - the latter makes use of `/etc/ld.so.conf.d`. That results in the amdgpu folders being filtered out from the `BUILD.bazel` generation. I think that's a better solution than my first attempt.

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-:-N-Adq7_5zxBJXCMgJd-:b6rcjna)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @EricCousineau-TRI and @sloretz)

*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-AOULtCJkJWGnDnTk7:-N-Af4GK66YkFSxC1qPK:bm8gguv) (no related file):*

nit At present, I believe only you + Calder can repro this.

Can I ask you to post a draft PR, similar to #59, that shows the repro? I tried on my side w/ your suggestion, but couldn't capture it.
</blockquote></details>

Would a script building and running a singularity container work?

---

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/61)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @sloretz)

*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-AOULtCJkJWGnDnTk7:-N-AknjU7BS1ObhCVp0P:b-p1lcvd) (no related file):*

Would a script building and running a singularity container work?
</blockquote></details>

Yup! Whatever's best for ya.

---

sloretz:
*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-AOULtCJkJWGnDnTk7:-N-AvQ0W1SO4rYY8WjPQ:bdbjc5g) (no related file):*

Yup! Whatever's best for ya.
</blockquote></details>

Reproduction script here: https://github.com/RobotLocomotion/drake-ros/pull/59#issuecomment-1093559724

---

EricCousineau-TRI (commented):
> Reviewed 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/61)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @sloretz)

*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-AOULtCJkJWGnDnTk7:-N-K95sI5W_-hch2c38c:bwrc7li) (no related file):*

Reproduction script here: https://github.com/RobotLocomotion/drake-ros/pull/59#issuecomment-1093559724
</blockquote></details>

OK Can confirm repro!

---

EricCousineau-TRI (changes requested):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/61)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @sloretz)

*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-K9A-XEa00nKq9q5zg:-N-K9A-XEa00nKq9q5zh:b-xja2b3) (no related file):*
CI is currently failing; perhaps filtering was too aggressive?
https://github.com/RobotLocomotion/drake-ros/runs/5893249247?check_suite_focus=true#step:6:40
```
[106 / 118] Compiling ros2_example_apps/oracle.cc; 7s processwrapper-sandbox ... (2 actions, 1 running)
ERROR: /__w/drake-ros/drake-ros/ros2_example_bazel_installed/ros2_example_apps/BUILD.bazel:20:14: Linking ros2_example_apps/_oracle_cc_shimmed failed: (Exit 1): gcc failed: error executing command /usr/bin/gcc @bazel-out/k8-opt/bin/ros2_example_apps/_oracle_cc_shimmed-2.params

Use --sandbox_debug to see verbose messages from the sandbox
bazel-out/k8-opt/bin/_solib_k8/_U_S_Sros2_Uexample_Ucommon_Cros2_Uexample_Ucommon_Umsgs_Ucc___Uros2_Uexample_Ucommon/libros2_example_common_msgs__rosidl_typesupport_fastrtps_cpp.so: error: undefined reference to 'eprosima::fastcdr::Cdr::serialize(char)'
bazel-out/k8-opt/bin/_solib_k8/_U_S_Sros2_Uexample_Ucommon_Cros2_Uexample_Ucommon_Umsgs_Ucc___Uros2_Uexample_Ucommon/libros2_example_common_msgs__rosidl_typesupport_fastrtps_cpp.so: error: undefined reference to 'eprosima::fastcdr::Cdr::deserialize(char&)'
```

---

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/61)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @sloretz)

*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-AOxJw9jkhBgjcZNGt:-N-KFm-L-fFWI_lCzbRN:bljeueu) (no related file):*

Per above, we should have a reproducible confirmation that we can read libraries listed for `/etc/ld.so.conf.d`.
(I believe they should be available, so hopefully no extra work is needed.)
</blockquote></details>

OK After thinking about this, `bazel test` on binaries that are dynamically linking this (hopefully any / all?) should be sufficient testing.

---
___
*[bazel_ros2_rules/ros2/ros2bzl/scraping/system.py, line 33 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-KEqKz3oOoslM-Rt9j:-N-KEqKz3oOoslM-Rt9k:b-7t7o4u) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/25e4a675684d7ed15e5c7f3b50b42a403062ac93/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py#L33)):*
> ```Python
>     """Get standard search paths for the linker."""
>     link_dirs = set(os.environ.get('LIBRARY_PATH', '').split(':'))
>     output = subprocess.run(
> ```

nit Why not just lump both "link" + "shared lib" for now, and make parsing a bit simpler? e.g. in addition to `LIBRARY_PATH` and `LD_LIBRARY_PATH`, just add these?
```
lines = shell("cat /etc/ld.so.conf.d/*.conf").splitlines()
for line in lines:
   line = line.split('#')[0].strip()
   if isdir(line):
      dirs.add(line)
```

It's probably overfit to Debian, but :shrug: for now?
(unless your current approach is better for Mac?)
___
*[bazel_ros2_rules/ros2/ros2bzl/scraping/system.py, line 53 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-KEPGPCx63Qa75mn32:-N-KEPGPCx63Qa75mn33:byi5yha) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/25e4a675684d7ed15e5c7f3b50b42a403062ac93/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py#L53)):*
> ```Python
>     output = subprocess.run(
>         ['ldconfig', '--verbose'],
>         check=False,  # TODO(sloretz) why does this proces have a non-zero exit
> ```

FYI On my system:
```
$ cd ~
$ ldconfig --verbose
...
/sbin/ldconfig.real: Can't create temporary cache file /etc/ld.so.cache~: Permission denied
````

**Review by sloretz — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/61)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @EricCousineau-TRI and @sloretz)

___
*[bazel_ros2_rules/ros2/ros2bzl/scraping/system.py, line 33 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-KEqKz3oOoslM-Rt9j:-N-OeTjZ7xMQGR1oSHDn:bql2a93) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/25e4a675684d7ed15e5c7f3b50b42a403062ac93/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py#L33)):*
>  in addition to LIBRARY_PATH and LD_LIBRARY_PATH

And reading the linker script via `ld --verbose`, or are you saying we don't need to read it?

> and make parsing a bit simpler? e.g. [...] just add these?

I think getting the output from `ldconfig` is less risky. If this code misses a directory that will be searched by the dynamic loader, and the directory contains a library used by one of the ROS 2 libraries in the archive, then the `BUILD.bazel` will be invalid.

I think more than `dirs.add(line)` would be necessary because we can't know if someone will install a package that uses more features of these config files. It looks like `ldconfig` reads `/etc/ld.so.conf.d` because there's an `include` directive in the config file `/etc/ld.so.conf`, but there could be more configuration in `/etc/ld.so.conf`. The content of the [config files supports glob patterns](https://github.com/lattera/glibc/blob/895ef79e04a953cac1493863bcae29ad85657ee1/elf/ldconfig.c#L1216), so `line` may need to be expanded. The [config file parsing is recursive](https://github.com/lattera/glibc/blob/895ef79e04a953cac1493863bcae29ad85657ee1/elf/ldconfig.c#L1222), so a file in `/etc/ld.so.conf.d` could itself have another `include` directive. It also [hard codes some system directories](https://github.com/lattera/glibc/blob/895ef79e04a953cac1493863bcae29ad85657ee1/elf/ldconfig.c#L1385-L1387). I recommend we use `ldconfig` instead.
___
*[bazel_ros2_rules/ros2/ros2bzl/scraping/system.py, line 53 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-KEPGPCx63Qa75mn32:-N-OqhZHEfIGEw06aBuN:b-9iarpm) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/25e4a675684d7ed15e5c7f3b50b42a403062ac93/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py#L53)):*

FYI On my system:
```
$ cd ~
$ ldconfig --verbose
...
/sbin/ldconfig.real: Can't create temporary cache file /etc/ld.so.cache~: Permission denied
````
</blockquote></details>

Do you see the same output with `ldconfig -XN --verbose`?

**Review by sloretz — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/61)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @EricCousineau-TRI and @sloretz)

*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-AOxJw9jkhBgjcZNGt:-N-P1VLODdHhJXSka_Zk:b-7df1yu) (no related file):*
> OK After thinking about this, bazel test on binaries that are dynamically linking this (hopefully any / all?) should be sufficient testing.

Are you saying you'd like a test added to catch this in the future? If so, what do you mean by `this` in "dynamically linking this"?

---
*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-K9A-XEa00nKq9q5zg:-N-P1K7c0GFw47g90-es:b-v8wb7w) (no related file):*

CI is currently failing; perhaps filtering was too aggressive?
https://github.com/RobotLocomotion/drake-ros/runs/5893249247?check_suite_focus=true#step:6:40
```
[106 / 118] Compiling ros2_example_apps/oracle.cc; 7s processwrapper-sandbox ... (2 actions, 1 running)
ERROR: /__w/drake-ros/drake-ros/ros2_example_bazel_installed/ros2_example_apps/BUILD.bazel:20:14: Linking ros2_example_apps/_oracle_cc_shimmed failed: (Exit 1): gcc failed: error executing command /usr/bin/gcc @bazel-out/k8-opt/bin/ros2_example_apps/_oracle_cc_shimmed-2.params

Use --sandbox_debug to see verbose messages from the sandbox
bazel-out/k8-opt/bin/_solib_k8/_U_S_Sros2_Uexample_Ucommon_Cros2_Uexample_Ucommon_Umsgs_Ucc___Uros2_Uexample_Ucommon/libros2_example_common_msgs__rosidl_typesupport_fastrtps_cpp.so: error: undefined reference to 'eprosima::fastcdr::Cdr::serialize(char)'
bazel-out/k8-opt/bin/_solib_k8/_U_S_Sros2_Uexample_Ucommon_Cros2_Uexample_Ucommon_Umsgs_Ucc___Uros2_Uexample_Ucommon/libros2_example_common_msgs__rosidl_typesupport_fastrtps_cpp.so: error: undefined reference to 'eprosima::fastcdr::Cdr::deserialize(char&)'
```
</blockquote></details>

Ah, yeah it shouldn't be treating `LD_LIBRARY_PATH` as system library directories. Fixed in https://github.com/RobotLocomotion/drake-ros/pull/61/commits/a6b13675c6a0f9900983c9780264cb3a68226af9

---

**Review by EricCousineau-TRI — CHANGES_REQUESTED** [review]
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/61)* status: all files reviewed, 3 unresolved discussions (waiting on @EricCousineau-TRI and @sloretz)

*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-AOxJw9jkhBgjcZNGt:-N-PpuUY6DZD_EOXoU-i:b-2m03) (no related file):*

> OK After thinking about this, bazel test on binaries that are dynamically linking this (hopefully any / all?) should be sufficient testing.

Are you saying you'd like a test added to catch this in the future? If so, what do you mean by `this` in "dynamically linking this"?
</blockquote></details>

Sorry for being unclear - I was wondering if we should explicitly test that `bazel test` permits running a binary which dynamically links a shared library outside of the sandbox.

However, I redacted that request because I think the test passing itself should be evidence of it working.

Does that make sense?

---
*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-Pq95nEGMDpkK8lYMG:-N-Pq95nEGMDpkK8lYMH:b-jq793q) (no related file):*
Working: Given functionality change, re-testing.

---
___
*[bazel_ros2_rules/ros2/ros2bzl/scraping/system.py, line 33 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-KEqKz3oOoslM-Rt9j:-N-Pownq9q478JHJe4UI:b4xueq1) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/25e4a675684d7ed15e5c7f3b50b42a403062ac93/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py#L33)):*

>  in addition to LIBRARY_PATH and LD_LIBRARY_PATH

And reading the linker script via `ld --verbose`, or are you saying we don't need to read it?

> and make parsing a bit simpler? e.g. [...] just add these?

I think getting the output from `ldconfig` is less risky. If this code misses a directory that will be searched by the dynamic loader, and the directory contains a library used by one of the ROS 2 libraries in the archive, then the `BUILD.bazel` will be invalid.

I think more than `dirs.add(line)` would be necessary because we can't know if someone will install a package that uses more features of these config files. It looks like `ldconfig` reads `/etc/ld.so.conf.d` because there's an `include` directive in the config file `/etc/ld.so.conf`, but there could be more configuration in `/etc/ld.so.conf`. The content of the [config files supports glob patterns](https://github.com/lattera/glibc/blob/895ef79e04a953cac1493863bcae29ad85657ee1/elf/ldconfig.c#L1216), so `line` may need to be expanded. The [config file parsing is recursive](https://github.com/lattera/glibc/blob/895ef79e04a953cac1493863bcae29ad85657ee1/elf/ldconfig.c#L1222), so a file in `/etc/ld.so.conf.d` could itself have another `include` directive. It also [hard codes some system directories](https://github.com/lattera/glibc/blob/895ef79e04a953cac1493863bcae29ad85657ee1/elf/ldconfig.c#L1385-L1387). I recommend we use `ldconfig` instead.
</blockquote></details>

OK Ah, wasn't aware of that level of configurability - sounds good to me, then!
___
*[bazel_ros2_rules/ros2/ros2bzl/scraping/system.py, line 53 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-KEPGPCx63Qa75mn32:-N-Pp6xWABGnJ1Uh8yvk:b-72fam0) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/25e4a675684d7ed15e5c7f3b50b42a403062ac93/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py#L53)):*

Do you see the same output with `ldconfig -XN --verbose`?
</blockquote></details>

Nope, that fixes it, and now it returns 0.
Can you remove the `TODO` and restore `check=True`?
___
*[bazel_ros2_rules/ros2/ros2bzl/scraping/system.py, line 98 at r3](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-PpHvf2G2dmnH79MB2:-N-PpHvf2G2dmnH79MB3:b-ksrdw4) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/2448ad07ee1a7bb5b772f077d881a73356b70efb/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py#L98)):*
> ```Python
>     if link_flags:
>         cmd.extend(link_flags)
>     for path in tuple(set(os.environ.get('LIBRARY_PATH', '').split(':'))):
> ```

nit Unclear why `tuple(set())` is used; I think `set()` will be sufficient?

(also, consider assigning this expression to separate variable, e.g. `library_dirs = set(...)`)

**Review by EricCousineau-TRI — CHANGES_REQUESTED** [review]
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/61)* status: all files reviewed, 3 unresolved discussions (waiting on @EricCousineau-TRI and @sloretz)

*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-Pq95nEGMDpkK8lYMG:-N-Psx5EDZ1-NckviXN8:blq4g8q) (no related file):*

Working: Given functionality change, re-testing.
</blockquote></details>

This is currently failing for me on 7d63bb4 from #59 (which uses r3 of this PR); it did not fail on the prior revision of that commit, 74895c2
```
ERROR: Failed to setup @ros2 repository: './run.bash {output_dir}/external/bazel_ros2_rules/ros2/generate_build_file.py -s {output_dir}/external/ros2/archive:archive -d distro_metadata.json ros2 -o BUILD.bazel' exited with 1
--- captured stderr ---
Traceback (most recent call last):
  File "{output_dir}/external/bazel_ros2_rules/ros2/generate_build_file.py", line 205, in <module>
    main()
  File "{output_dir}/external/bazel_ros2_rules/ros2/generate_build_file.py", line 199, in main
    write_build_file(
  File "{output_dir}/external/bazel_ros2_rules/ros2/generate_build_file.py", line 134, in write_build_file
    _, template, config = configure_package_cc_library(
  File "{output_dir}/external/bazel_ros2_rules/ros2/ros2bzl/templates.py", line 108, in configure_package_cc_library
    data.extend(
  File "{output_dir}/external/bazel_ros2_rules/ros2/ros2bzl/templates.py", line 109, in <genexpr>
    sandbox(find_library_path(library))
  File "{output_dir}/external/bazel_ros2_rules/ros2/ros2bzl/sandboxing.py", line 25, in _map
    path = os.path.normpath(path)
  File "/usr/lib/python3.8/posixpath.py", line 336, in normpath
    path = os.fspath(path)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

Can you take a look?

---

**Review by sloretz — COMMENTED** [review]
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/61)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @EricCousineau-TRI and @sloretz)

*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-AOxJw9jkhBgjcZNGt:-N-PyJVq3FmaiE9tLtXR:b-5thqby) (no related file):*

Sorry for being unclear - I was wondering if we should explicitly test that `bazel test` permits running a binary which dynamically links a shared library outside of the sandbox.

However, I redacted that request because I think the test passing itself should be evidence of it working.

Does that make sense?
</blockquote></details>

That makes sense to me.

---
*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-Pq95nEGMDpkK8lYMG:-N-Q0D0v1z48g-c_vUBB:b-mcdcng) (no related file):*

This is currently failing for me on 7d63bb4 from #59 (which uses r3 of this PR); it did not fail on the prior revision of that commit, 74895c2
```
ERROR: Failed to setup @ros2 repository: './run.bash {output_dir}/external/bazel_ros2_rules/ros2/generate_build_file.py -s {output_dir}/external/ros2/archive:archive -d distro_metadata.json ros2 -o BUILD.bazel' exited with 1
--- captured stderr ---
Traceback (most recent call last):
  File "{output_dir}/external/bazel_ros2_rules/ros2/generate_build_file.py", line 205, in <module>
    main()
  File "{output_dir}/external/bazel_ros2_rules/ros2/generate_build_file.py", line 199, in main
    write_build_file(
  File "{output_dir}/external/bazel_ros2_rules/ros2/generate_build_file.py", line 134, in write_build_file
    _, template, config = configure_package_cc_library(
  File "{output_dir}/external/bazel_ros2_rules/ros2/ros2bzl/templates.py", line 108, in configure_package_cc_library
    data.extend(
  File "{output_dir}/external/bazel_ros2_rules/ros2/ros2bzl/templates.py", line 109, in <genexpr>
    sandbox(find_library_path(library))
  File "{output_dir}/external/bazel_ros2_rules/ros2/ros2bzl/sandboxing.py", line 25, in _map
    path = os.path.normpath(path)
  File "/usr/lib/python3.8/posixpath.py", line 336, in normpath
    path = os.fspath(path)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

Can you take a look?
</blockquote></details>

It looks like the scraping system is using `find_library_path()` to look for plugins and runtime  libraries, so 224bb9b adds those back as search paths.

---
*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-QDL_l5JB57_QbwnSR:-N-QDL_l5JB57_QbwnSS:brp0amc) (no related file):*
I'm seeing new failures to build when an existing ROS installation is sourced. I'm looking into why the libraries are found in `/opt/ros/rolling` instead of in the archive.


```console
$ bazel build //...
ERROR: /home/osrf/.cache/bazel/_bazel_osrf/383f079a2635d21e3b905b3ffc908f7b/external/ros2/BUILD.bazel:1792:11: @ros2//:_libstatistics_collector_py_c: invalid label '/opt/ros/rolling/lib/python3.8/site-packages/libstatistics_collector/liblibstatistics_collector_test_msgs__rosidl_generator_py.so' in element 0 of attribute 'srcs' in 'cc_library' rule: invalid target name '/opt/ros/rolling/lib/python3.8/site-packages/libstatistics_collector/liblibstatistics_collector_test_msgs__rosidl_generator_py.so': target names may not start with '/'
ERROR: /home/osrf/.cache/bazel/_bazel_osrf/383f079a2635d21e3b905b3ffc908f7b/external/ros2/BUILD.bazel:1792:11: @ros2//:_libstatistics_collector_py_c: invalid label '/opt/ros/rolling/lib/python3.8/site-packages/libstatistics_collector/liblibstatistics_collector_test_msgs__rosidl_generator_py.so' in element 0 of attribute 'data' in 'cc_library' rule: invalid target name '/opt/ros/rolling/lib/python3.8/site-packages/libstatistics_collector/liblibstatistics_collector_test_msgs__rosidl_generator_py.so': target names may not start with '/'
ERROR: Traceback (most recent call last):
	File "/home/osrf/.cache/bazel/_bazel_osrf/383f079a2635d21e3b905b3ffc908f7b/external/ros2/BUILD.bazel", line 1801, column 16, in <toplevel>
		srcs = glob(["{}/**/*.py".format(x) for x in ["/opt/ros/rolling/lib/python3.8/site-packages/libstatistics_collector"]]),
Error in glob: pattern cannot be absolute
ERROR: /home/osrf/ws/drake/src/drake-ros/ros2_example_bazel_installed/ros2_example_apps/BUILD.bazel:9:24: Target '@ros2//:builtin_interfaces_defs' contains an error and its package is in error and referenced by '//ros2_example_apps:_ros2_example_apps_msgs__rosidl_typesupport_introspection_cpp_gen'
ERROR: /home/osrf/ws/drake/src/drake-ros/ros2_example_bazel_installed/ros2_example_apps/BUILD.bazel:9:24: every rule of type rosidl_generate_genrule implicitly depends upon the target '@ros2//:rosidl', but this target could not be found because of: no such target '@ros2//:rosidl': target 'rosidl' not declared in package '' defined by /home/osrf/.cache/bazel/_bazel_osrf/383f079a2635d21e3b905b3ffc908f7b/external/ros2/BUILD.bazel
ERROR: Analysis of target '//ros2_example_apps:_ros2_example_apps_msgs__rosidl_typesupport_introspection_cpp_gen' failed; build aborted: Analysis failed
INFO: Elapsed time: 0.107s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded, 0 targets configured)
```

---
___
*[bazel_ros2_rules/ros2/ros2bzl/scraping/system.py, line 53 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-KEPGPCx63Qa75mn32:-N-PyFyI8rG2aRfjjC53:btdx5ax) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/25e4a675684d7ed15e5c7f3b50b42a403062ac93/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py#L53)):*

Nope, that fixes it, and now it returns 0.
Can you remove the `TODO` and restore `check=True`?
</blockquote></details>

Removed the TOD and restored the check in 4a1a5e5b47867dc004d2830f3c0b9144b3b6bf66
___
*[bazel_ros2_rules/ros2/ros2bzl/scraping/system.py, line 98 at r3](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-PpHvf2G2dmnH79MB2:-N-PyVYB9YjOpJZI0V8T:bus0cwg) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/2448ad07ee1a7bb5b772f077d881a73356b70efb/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py#L98)):*

nit Unclear why `tuple(set())` is used; I think `set()` will be sufficient?

(also, consider assigning this expression to separate variable, e.g. `library_dirs = set(...)`)
</blockquote></details>

It was left over from having it as a global variable. I wanted it to be immutable if it was going to be global. 36bbb2d

EricCousineau-TRI (changes requested):
> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/61)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @EricCousineau-TRI and @sloretz)

*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-Pq95nEGMDpkK8lYMG:-N-TnfilC4shSm5SrCK9:b-jmaaem) (no related file):*

It looks like the scraping system is using `find_library_path()` to look for plugins and runtime  libraries, so 224bb9b adds those back as search paths.
</blockquote></details>

Thanks! Confirmed fixed in 9ccb02d

Will leave this open pending add'l functional changes.

---

sloretz (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-:-N-U69u69B_QrrRX6WBd:bnfp4nl)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @EricCousineau-TRI)

*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-QDL_l5JB57_QbwnSR:-N-U420p4QuBoYsJQQY6:b-vkiv3h) (no related file):*

I'm seeing new failures to build when an existing ROS installation is sourced. I'm looking into why the libraries are found in `/opt/ros/rolling` instead of in the archive.


```console
$ bazel build //...
ERROR: /home/osrf/.cache/bazel/_bazel_osrf/383f079a2635d21e3b905b3ffc908f7b/external/ros2/BUILD.bazel:1792:11: @ros2//:_libstatistics_collector_py_c: invalid label '/opt/ros/rolling/lib/python3.8/site-packages/libstatistics_collector/liblibstatistics_collector_test_msgs__rosidl_generator_py.so' in element 0 of attribute 'srcs' in 'cc_library' rule: invalid target name '/opt/ros/rolling/lib/python3.8/site-packages/libstatistics_collector/liblibstatistics_collector_test_msgs__rosidl_generator_py.so': target names may not start with '/'
ERROR: /home/osrf/.cache/bazel/_bazel_osrf/383f079a2635d21e3b905b3ffc908f7b/external/ros2/BUILD.bazel:1792:11: @ros2//:_libstatistics_collector_py_c: invalid label '/opt/ros/rolling/lib/python3.8/site-packages/libstatistics_collector/liblibstatistics_collector_test_msgs__rosidl_generator_py.so' in element 0 of attribute 'data' in 'cc_library' rule: invalid target name '/opt/ros/rolling/lib/python3.8/site-packages/libstatistics_collector/liblibstatistics_collector_test_msgs__rosidl_generator_py.so': target names may not start with '/'
ERROR: Traceback (most recent call last):
	File "/home/osrf/.cache/bazel/_bazel_osrf/383f079a2635d21e3b905b3ffc908f7b/external/ros2/BUILD.bazel", line 1801, column 16, in <toplevel>
		srcs = glob(["{}/**/*.py".format(x) for x in ["/opt/ros/rolling/lib/python3.8/site-packages/libstatistics_collector"]]),
Error in glob: pattern cannot be absolute
ERROR: /home/osrf/ws/drake/src/drake-ros/ros2_example_bazel_installed/ros2_example_apps/BUILD.bazel:9:24: Target '@ros2//:builtin_interfaces_defs' contains an error and its package is in error and referenced by '//ros2_example_apps:_ros2_example_apps_msgs__rosidl_typesupport_introspection_cpp_gen'
ERROR: /home/osrf/ws/drake/src/drake-ros/ros2_example_bazel_installed/ros2_example_apps/BUILD.bazel:9:24: every rule of type rosidl_generate_genrule implicitly depends upon the target '@ros2//:rosidl', but this target could not be found because of: no such target '@ros2//:rosidl': target 'rosidl' not declared in package '' defined by /home/osrf/.cache/bazel/_bazel_osrf/383f079a2635d21e3b905b3ffc908f7b/external/ros2/BUILD.bazel
ERROR: Analysis of target '//ros2_example_apps:_ros2_example_apps_msgs__rosidl_typesupport_introspection_cpp_gen' failed; build aborted: Analysis failed
INFO: Elapsed time: 0.107s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded, 0 targets configured)
```
</blockquote></details>

I've confirmed this is an issue on develop, so I don't think it needs to block this PR. I opened an issue : https://github.com/RobotLocomotion/drake-ros/issues/63

---
___
*[bazel_ros2_rules/ros2/ros2bzl/scraping/system.py, line 43 at r4](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-TAeyj19Hf2nFMKLT2:-N-U6DnN8K4pwF_yYj3P:b4rkf50) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/224bb9b7d9c9e1f277470b47d804eed4ea4cd4b6/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py#L43)):*

nit Spelling, "emtpy"
</blockquote></details>

Fixed spelling in c0ddbfaf00a0624b360508077c75f8c0f007b48d
___
*[bazel_ros2_rules/ros2/ros2bzl/scraping/system.py, line 61 at r4](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-TAha60n2f9m3EeLf3:-N-U6F9v0bIJO3YM7Alv:b4rkf50) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/224bb9b7d9c9e1f277470b47d804eed4ea4cd4b6/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py#L61)):*

nit Spelling, "emtpy"
</blockquote></details>

Fixed spelling in c0ddbfaf00a0624b360508077c75f8c0f007b48d
___
*[bazel_ros2_rules/ros2/ros2bzl/scraping/system.py, line 97 at r4](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-TAjb967NPvvIi4D87:-N-U6Gse2d5fCjHCR9cN:b-1m4tv2) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/224bb9b7d9c9e1f277470b47d804eed4ea4cd4b6/bazel_ros2_rules/ros2/ros2bzl/scraping/system.py#L97)):*

nit There are four occurrences of `for path in paths: cmd.extend(["-L", path])`

Consider aggregating into a single `library_paths` variable, and using one for loop.
</blockquote></details>

consolidated loops in d8ff92623ed9e39d916b53ce18cde2e6bbbf345e

EricCousineau-TRI (commented):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/61)* status: all files reviewed, 1 unresolved discussion (waiting on @EricCousineau-TRI)

EricCousineau-TRI (approved):
@sloretz I prefer this approach over #65 - would you be OK if we merged it and re-examined the failures you encountered using `--nv` via Singularity?

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-:-N-csODLETgsfYgimkaj:bw9x9p4)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sloretz)

*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-N-Pq95nEGMDpkK8lYMG:-N-csLH-Aerqh_CZ7kji:byn5gto) (no related file):*

Thanks! Confirmed fixed in 9ccb02d

Will leave this open pending add'l functional changes.
</blockquote></details>

OK No further functional changes. I still think this is good.

---

EricCousineau-TRI (approved):
* it => this PR

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-:-N-cs_wD9xSxAudFKrIR:b4brbjz)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sloretz)

sloretz (commented):
Yeah, I'm fine with this PR being merged. I'll think more about how to handle other entries in LD_LIBRARY_PATH not set by the ROS workspace.

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/61#-:-N-dBBWH1rHNHa124Ryc:b-eeym)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sloretz)

EricCousineau-TRI:
Sweet, thank you!

PR #93, opened 2022-07-21
Add introspection of Drake output values as ROS topics

gbiggs:
This PR adds the `drake_ros_introspection` package, which exposes selected Drake system ports as ROS 2 topics.

Most of this work was done by @hidmic in #32. I cleaned up the API a little and added an extra example using the `ManipulationStation` example system.

The package currently has two shortcomings:
1. It does not integrate with `drake_ros_core`, which it should.
2. It only handles data going from Drake to ROS. It is not yet possible to take data from a ROS topic and push it into a Drake system.

Nevertheless I'd like to get this merged then address these two issues as separate PRs.

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/93)


EricCousineau-TRI:
Currently paused, closing for now. Tracked via #228

Issue #105, opened 2022-08-02
ros_cc: Cannot run binaries using Rlocation via bazel-bin


EricCousineau-TRI:
In Anzu, I was trying to write a test that I could run outside of `bazel run` and `bazel test` (we must be able to do this).
However, I got errors about Drake not being able to find some basic files.

Ends up, it's an issue with how `drake-ros` does it's shimming in C++.

Filing PR that has sufficient testing.

EricCousineau-TRI:
See #106

EricCousineau-TRI:
Fixed via #106

PR #106, opened 2022-08-02
Add testing for ros_cc targets that use runfiles

EricCousineau-TRI:
Towards #105

---
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/106)


EricCousineau-TRI:
@sloretz Can I ask you to fix / delegate this? I'd like this to ideally work in Anzu before we merge to `master`.
\cc @IanTheEngineer

EricCousineau-TRI:
My suggestion is, for the C++ shim, consider setting `RUNFILES_MANIFEST_FILE` or `RUNFILES_DIR`, if it's possible to somehow back that out from `/proc/self/exe`

bazel_ros2_rules/ros2/tools/dload_cc.bzl
```diff
@@ -30,8 +30,12 @@ using bazel::tools::cpp::runfiles::Runfiles;
 int main(int argc, const char * argv[]) {{
   std::string error;
   std::unique_ptr<Runfiles> runfiles(Runfiles::Create(argv[0], &error));
+  if (!runfiles && std::filesystem::is_symlink(argv[0])) {{
+    runfiles.reset(
+      Runfiles::Create(std::filesystem::read_symlink(argv[0]), &error));
+  }}
```
sloretz, line 36: @EricCousineau-TRI - this makes the test pass. Does it seem like the right fix to you?

It seems like the shim was unable to read its runfiles during `bazel run //:runfiles_direct_test` because `argv[0]` was a symlink. Resolving that symlink leads to the real executable adjacent to a `*.runfiles` directory and `*.runfiles_manifest` file.

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/106)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @sloretz)

___
*[`bazel_ros2_rules/ros2/tools/dload_cc.bzl` line 36 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8_bPW5fCEt1zz9EX9q-r2-36:-N8_e7DL4Z6trkx3H2mB:b2meb1o) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/b24e7bea7ec2eb285bba75fdb70b71d4cb66fcb8/bazel_ros2_rules/ros2/tools/dload_cc.bzl#L36)):*

@EricCousineau-TRI - this makes the test pass. Does it seem like the right fix to you?

It seems like the shim was unable to read its runfiles during `bazel run //:runfiles_direct_test` because `argv[0]` was a symlink. Resolving that symlink leads to the real executable adjacent to a `*.runfiles` directory and `*.runfiles_manifest` file.
</blockquote></details>

This still fails locally for me, so I don't think so.
I believe it has to do with the indirection of the wrapper binary, not the indirection via symlink. Basically, the wrapper binary does not have runfiles locatable via `/proc/self/exe`. This relates to what I pointed out here in Anzu PR:
https://github.com/EricCousineau-TRI/repro/commit/f62a463a8973

Does this make sense to you?
___
*[`ros2_example_bazel_installed/test/runfiles_test.cc` line 32 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8_eLGNDIh5JwpMc5Pa:-N8_eLGNDIh5JwpMc5Pb:b-y4bi2v) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/b24e7bea7ec2eb285bba75fdb70b71d4cb66fcb8/ros2_example_bazel_installed/test/runfiles_test.cc#L32)):*
> ```cpp
>   } else {
>     // This mode fails.
>     const std::string& argv0 = std::filesystem::read_symlink({
> ```

FYI This should resolve the symlink, so I don't think your change fixes this just yet.

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/106)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @sloretz)

___
*[`bazel_ros2_rules/ros2/tools/dload_cc.bzl` line 36 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8_bPW5fCEt1zz9EX9q-r2-36:-N8_fA-X5E4gibx1mL8-:bxttf6n) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/b24e7bea7ec2eb285bba75fdb70b71d4cb66fcb8/bazel_ros2_rules/ros2/tools/dload_cc.bzl#L36)):*

This still fails locally for me, so I don't think so.
I believe it has to do with the indirection of the wrapper binary, not the indirection via symlink. Basically, the wrapper binary does not have runfiles locatable via `/proc/self/exe`. This relates to what I pointed out here in Anzu PR:
https://github.com/EricCousineau-TRI/repro/commit/f62a463a8973

Does this make sense to you?
</blockquote></details>

Ah, even worse, CI passes because it uses `bazel build //...` and then `bazel test //...`. I can reproduce this locally.

That is **bad**. Running a single test should not depend on the user building other tests; all things declared via `data` and `deps` should be accessible directly for a properly written Bazel binary / test, and we need `drake-ros` to be proper.

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/106)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @sloretz)

___
*[`ros2_example_bazel_installed/test/runfiles_direct_test.py` line 21 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8_foB9DfTXi8fz9xQt:-N8_foB9DfTXi8fz9xQu:bti5yr7) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/b24e7bea7ec2eb285bba75fdb70b71d4cb66fcb8/ros2_example_bazel_installed/test/runfiles_direct_test.py#L21)):*
> ```Python
>     minimal_env = {}
>     for key in keys:
>         if key in os.environ:
> ```

This logic is both confusing (redundant inner loop), and defeats the purpose of the test - we want to simulate a user running the binary outside of `bazel run` and `bazel test`, meaning we need to strip out the `RUNFILES*` and `TEST*` variables that Bazel injects.

does that make sense?
(Soz to be pedantic, but I believe it's very important to understand these nuances)

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/106)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @EricCousineau-TRI and @sloretz)

*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8_gmgK6VnUfsrc0rUT:-N8_gmgK6VnUfsrc0rUU:bch0hck) (no related file):*
Working: We should add (back?) `tags = ["manual"]` to the generated shims.

___
___
*[`bazel_ros2_rules/ros2/tools/dload_cc.bzl` line 36 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8_bPW5fCEt1zz9EX9q-r2-36:-N8_gjS75nqufixFV8GM:b-mv50lq) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/b24e7bea7ec2eb285bba75fdb70b71d4cb66fcb8/bazel_ros2_rules/ros2/tools/dload_cc.bzl#L36)):*

Ah, even worse, CI passes because it uses `bazel build //...` and then `bazel test //...`. I can reproduce this locally.

That is **bad**. Running a single test should not depend on the user building other tests; all things declared via `data` and `deps` should be accessible directly for a properly written Bazel binary / test, and we need `drake-ros` to be proper.
</blockquote></details>

Working: Seeing if I can stab at a fix.

jwnimmer-tri:
*[`ros2_example_bazel_installed/test/runfiles_direct_test.py` line 21 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8_foB9DfTXi8fz9xQt:-N8_gWrY9W3Sf5coGnL5:b-8zqx7f) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/b24e7bea7ec2eb285bba75fdb70b71d4cb66fcb8/ros2_example_bazel_installed/test/runfiles_direct_test.py#L21)):*

This logic is both confusing (redundant inner loop), and defeats the purpose of the test - we want to simulate a user running the binary outside of `bazel run` and `bazel test`, meaning we need to strip out the `RUNFILES*` and `TEST*` variables that Bazel injects.

does that make sense?
(Soz to be pedantic, but I believe it's very important to understand these nuances)
</blockquote></details>

I think it's impossible to correctly test "What happens when a user directly runs bazel-bin/foo/bar?" inside of a bazel-declared test.  If you want a test for that scenario, you'll need to write it as a shebang program that's called by your CI separately from `bazel test //...`.

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/106)* status: 0 of 6 files reviewed, 3 unresolved discussions (waiting on @EricCousineau-TRI, @jwnimmer-tri, and @sloretz)

___
*[`ros2_example_bazel_installed/test/runfiles_direct_test.py` line 21 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8_foB9DfTXi8fz9xQt:-N8_h2wj8Q6HPhrEAW3f:b-pa8z00) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/b24e7bea7ec2eb285bba75fdb70b71d4cb66fcb8/ros2_example_bazel_installed/test/runfiles_direct_test.py#L21)):*
<details><summary><i>Previously, jwnimmer-tri (Jeremy Nimmer) wrote…</i></summary><blockquote>

I think it's impossible to correctly test "What happens when a user directly runs bazel-bin/foo/bar?" inside of a bazel-declared test.  If you want a test for that scenario, you'll need to write it as a shebang program that's called by your CI separately from `bazel test //...`.
</blockquote></details>

Ah, fair! Since the recursive deps + data declarations will require some crafting, will file issue.
(unless there's an easy technique that you've found, beyond crafting `filegroup("source_tree_deps")`?)

For now, though, I think this test exercises at least this failure mode (from r1).

jwnimmer-tri:
*[`ros2_example_bazel_installed/test/runfiles_direct_test.py` line 21 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8_foB9DfTXi8fz9xQt:-N8_iaU23wybYuKP3r3d:b-jcqkic) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/b24e7bea7ec2eb285bba75fdb70b71d4cb66fcb8/ros2_example_bazel_installed/test/runfiles_direct_test.py#L21)):*

Ah, fair! Since the recursive deps + data declarations will require some crafting, will file issue.
(unless there's an easy technique that you've found, beyond crafting `filegroup("source_tree_deps")`?)

For now, though, I think this test exercises at least this failure mode (from r1).
</blockquote></details>

I haven't studied this test carefully, but I realized there is one exception to "impossible".  To test "what does bazel-bin/foo/bar look like?", you can write a within-bazel python test that:

(1) creates a scratch folder
(2) places an _entire source code tree_ there (WORKSPACE, BUILD files, source code, data, etc.)
(3) shells out to `subprocess.call("bazel build //...", cwd=tmpdir)` to do the build
(4) shells out to `subprocess.call("bazel-bin/foo/bar", cwd=tmpdir` to check what the user would ses.

In other words, you can't run the test on the same project as itself, but you can create a temporary dummy project and then test it there.

sloretz (commented):
> Reviewed 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/106)* status: 1 of 6 files reviewed, 3 unresolved discussions (waiting on @EricCousineau-TRI, @jwnimmer-tri, and @sloretz)

___
*[`ros2_example_bazel_installed/test/runfiles_direct_test.py` line 21 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8_foB9DfTXi8fz9xQt:-N8_iKfy8DpFqCTacuQn:bum39cp) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/b24e7bea7ec2eb285bba75fdb70b71d4cb66fcb8/ros2_example_bazel_installed/test/runfiles_direct_test.py#L21)):*
> This logic is both confusing (redundant inner loop), and defeats the purpose of the test - we want to simulate a user running the binary outside of bazel run and bazel test, meaning we need to strip out the RUNFILES* and TEST* variables that Bazel injects.


`minimal_env[key] = os.environ[key]` raises `KeyError` when `HOME` does not exist in `os.environ`, so this line skips it. All the other variables are still stripped out.

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/106)* status: 1 of 6 files reviewed, 3 unresolved discussions (waiting on @EricCousineau-TRI, @jwnimmer-tri, and @sloretz)

___
*[`ros2_example_bazel_installed/test/runfiles_direct_test.py` line 21 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8_foB9DfTXi8fz9xQt:-N8_jizz8Y2_pxjDqSe6:but9syr) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/b24e7bea7ec2eb285bba75fdb70b71d4cb66fcb8/ros2_example_bazel_installed/test/runfiles_direct_test.py#L21)):*
>  `raises KeyError`

derp! misread the `for` as an `if` :facepalm: 

> I haven't studied this test carefully, but I realized there is one exception to "impossible".  [...]

ah, that'd be nicer! will still try to fix as-is, then see if it's quick to separate-workspace testing strategy.

sloretz (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/106)* status: 1 of 6 files reviewed, 3 unresolved discussions (waiting on @EricCousineau-TRI, @jwnimmer-tri, and @sloretz)

___
*[`bazel_ros2_rules/ros2/tools/dload_cc.bzl` line 36 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8_bPW5fCEt1zz9EX9q-r2-36:-N8_kKCa1kIECGpfxGPN:b9dxx80) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/b24e7bea7ec2eb285bba75fdb70b71d4cb66fcb8/bazel_ros2_rules/ros2/tools/dload_cc.bzl#L36)):*
> This still fails locally for me, so I don't think so.

How do I reproduce the failure? `./bazel-bin/runfiles_direct_test` passes for me.

```
focal:osrf> bazel clean --expunge
INFO: Starting clean (this may take a while). Consider using --async if the clean takes more than several minutes.
focal:osrf> bazel build //...
Starting local Bazel server and connecting to it...
DEBUG: Rule 'ros2' indicated that a canonical reproducible form can be obtained by modifying arguments sha256 = "8835a0ece10ae720a4f054722f965df2fe0247e2ffd2f381a0e5152e43a85364"
DEBUG: Repository ros2 instantiated at:
  /home/osrf/ws/drake/src/drake-ros/ros2_example_bazel_installed/WORKSPACE:34:13: in <toplevel>
Repository rule ros2_archive defined at:
  /home/osrf/.cache/bazel/_bazel_osrf/383f079a2635d21e3b905b3ffc908f7b/external/bazel_ros2_rules/ros2/defs.bzl:307:31: in <toplevel>
INFO: Analyzed 71 targets (56 packages loaded, 7720 targets configured).
INFO: Found 71 targets...
INFO: Elapsed time: 28.195s, Critical Path: 11.09s
INFO: 519 processes: 423 internal, 96 linux-sandbox.
INFO: Build completed successfully, 519 total actions                  
focal:osrf> ./bazel-bin/runfiles_direct_test
/home/osrf/ws/drake/src/drake-ros/ros2_example_bazel_installed/test/runfiles_test_data.txt
Good!
/home/osrf/ws/drake/src/drake-ros/ros2_example_bazel_installed/test/runfiles_test_data.txt
Good!
```

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/106)* status: 1 of 6 files reviewed, 3 unresolved discussions (waiting on @EricCousineau-TRI, @jwnimmer-tri, and @sloretz)

___
*[`bazel_ros2_rules/ros2/tools/dload_cc.bzl` line 36 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8_bPW5fCEt1zz9EX9q-r2-36:-N8_kiaA40afsCIz06gi:b-kf9qvf) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/b24e7bea7ec2eb285bba75fdb70b71d4cb66fcb8/bazel_ros2_rules/ros2/tools/dload_cc.bzl#L36)):*

> This still fails locally for me, so I don't think so.

How do I reproduce the failure? `./bazel-bin/runfiles_direct_test` passes for me.

```
focal:osrf> bazel clean --expunge
INFO: Starting clean (this may take a while). Consider using --async if the clean takes more than several minutes.
focal:osrf> bazel build //...
Starting local Bazel server and connecting to it...
DEBUG: Rule 'ros2' indicated that a canonical reproducible form can be obtained by modifying arguments sha256 = "8835a0ece10ae720a4f054722f965df2fe0247e2ffd2f381a0e5152e43a85364"
DEBUG: Repository ros2 instantiated at:
  /home/osrf/ws/drake/src/drake-ros/ros2_example_bazel_installed/WORKSPACE:34:13: in <toplevel>
Repository rule ros2_archive defined at:
  /home/osrf/.cache/bazel/_bazel_osrf/383f079a2635d21e3b905b3ffc908f7b/external/bazel_ros2_rules/ros2/defs.bzl:307:31: in <toplevel>
INFO: Analyzed 71 targets (56 packages loaded, 7720 targets configured).
INFO: Found 71 targets...
INFO: Elapsed time: 28.195s, Critical Path: 11.09s
INFO: 519 processes: 423 internal, 96 linux-sandbox.
INFO: Build completed successfully, 519 total actions                  
focal:osrf> ./bazel-bin/runfiles_direct_test
/home/osrf/ws/drake/src/drake-ros/ros2_example_bazel_installed/test/runfiles_test_data.txt
Good!
/home/osrf/ws/drake/src/drake-ros/ros2_example_bazel_installed/test/runfiles_test_data.txt
Good!
```
</blockquote></details>

Ah, should've written that out.

To repro:
```
bazel clean --expunge --async
# don't build all targets
bazel run //:runfiles_direct_test
```

And sorry for not clarifying - it's not `./bazel-bin/runfiles_direct_test` that matters (it's a kinda poorly named meta test).
It's really ensuring that other tests will run directly.

Another more direct (less encapsulated) way is:
```
bazel clean --expunge --async
bazel build :runfiles_cc_test
./bazel-bin/runfiles_cc_test
```
___
*[`ros2_example_bazel_installed/test/runfiles_direct_test.py` line 21 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8_foB9DfTXi8fz9xQt:-N8_lamD7fu6zhwA_Ra4:b-oyu1x1) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/b24e7bea7ec2eb285bba75fdb70b71d4cb66fcb8/ros2_example_bazel_installed/test/runfiles_direct_test.py#L21)):*

>  `raises KeyError`

derp! misread the `for` as an `if` :facepalm: 

> I haven't studied this test carefully, but I realized there is one exception to "impossible".  [...]

ah, that'd be nicer! will still try to fix as-is, then see if it's quick to separate-workspace testing strategy.
</blockquote></details>

Ah, I could also just special-case our CI tests for repro above, which'd be more truthful pending proper source tree enumeration.

jwnimmer-tri:
*[`bazel_ros2_rules/ros2/tools/dload_cc.bzl` line 36 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8_bPW5fCEt1zz9EX9q-r2-36:-N8_m55ODWElTQ8hk_mD:b-rwsy3t) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/b24e7bea7ec2eb285bba75fdb70b71d4cb66fcb8/bazel_ros2_rules/ros2/tools/dload_cc.bzl#L36)):*

Ah, should've written that out.

To repro:
```
bazel clean --expunge --async
# don't build all targets
bazel run //:runfiles_direct_test
```

And sorry for not clarifying - it's not `./bazel-bin/runfiles_direct_test` that matters (it's a kinda poorly named meta test).
It's really ensuring that other tests will run directly.

Another more direct (less encapsulated) way is:
```
bazel clean --expunge --async
bazel build :runfiles_cc_test
./bazel-bin/runfiles_cc_test
```
</blockquote></details>

To make sure you're on the right track (and to see if I'm reading this correctly:

(1) Assuming you have an already-defined Bazel C++ binary named `:foo` (the "target").

(2) You want to generate a program `bar` that just (re)sets some environment variables and then does `evecv` to run `:foo` instead.

The correct way do do that is:

(a) The rule for "bar" has a `data` dependency on `:foo`.  Both foo and _all of foo's runfiles, transitively_ must also be runfiles of `:bar`.  Under the hood when the user runs the shim, then _never_ use `foo.runfiles` directory, they only ever use the `bar.runfiles` directly.

(b) In the shim, load the Runfiles object (let's call it `r`) based on bar's `argv[0]`.  (Or, if RUNFILES_DIR is already set, it means we're being called recursively, so use that environment variable instead of argv0.)

(c) In the shim, use `r` to look up the absolute path for the ":foo" resource path, and use that for what gets exec'd.

(d) In the shim, ask `r` for what `RUNFILES_MANIFEST_FILE` and/or `RUNFILES_DIR` should be set to, and set them to that, so that `foo` will obey those environment variable, and will not use its argv0 to find runfiles.

I do not believe that any other implementation approach is sound in all cases.

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/106)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @EricCousineau-TRI, @jwnimmer-tri, and @sloretz)

___
*[`bazel_ros2_rules/ros2/tools/dload_cc.bzl` line 36 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8_bPW5fCEt1zz9EX9q-r2-36:-N8_psX0CGtWYMdTzd-q:b-mqpk1k) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/b24e7bea7ec2eb285bba75fdb70b71d4cb66fcb8/bazel_ros2_rules/ros2/tools/dload_cc.bzl#L36)):*
<details><summary><i>Previously, jwnimmer-tri (Jeremy Nimmer) wrote…</i></summary><blockquote>

To make sure you're on the right track (and to see if I'm reading this correctly:

(1) Assuming you have an already-defined Bazel C++ binary named `:foo` (the "target").

(2) You want to generate a program `bar` that just (re)sets some environment variables and then does `evecv` to run `:foo` instead.

The correct way do do that is:

(a) The rule for "bar" has a `data` dependency on `:foo`.  Both foo and _all of foo's runfiles, transitively_ must also be runfiles of `:bar`.  Under the hood when the user runs the shim, then _never_ use `foo.runfiles` directory, they only ever use the `bar.runfiles` directly.

(b) In the shim, load the Runfiles object (let's call it `r`) based on bar's `argv[0]`.  (Or, if RUNFILES_DIR is already set, it means we're being called recursively, so use that environment variable instead of argv0.)

(c) In the shim, use `r` to look up the absolute path for the ":foo" resource path, and use that for what gets exec'd.

(d) In the shim, ask `r` for what `RUNFILES_MANIFEST_FILE` and/or `RUNFILES_DIR` should be set to, and set them to that, so that `foo` will obey those environment variable, and will not use its argv0 to find runfiles.

I do not believe that any other implementation approach is sound in all cases.
</blockquote></details>

Done. Yup yup, I believe that's what I just did here!
___
*[`ros2_example_bazel_installed/test/runfiles_direct_test.py` line 21 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8_foB9DfTXi8fz9xQt:-N8_pxEDBMcWbM-Eofu8:b-896fix) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/b24e7bea7ec2eb285bba75fdb70b71d4cb66fcb8/ros2_example_bazel_installed/test/runfiles_direct_test.py#L21)):*

Ah, I could also just special-case our CI tests for repro above, which'd be more truthful pending proper source tree enumeration.
</blockquote></details>

Done.

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/106)* status: 0 of 8 files reviewed, all discussions resolved (waiting on @jwnimmer-tri and @sloretz)

*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8_gmgK6VnUfsrc0rUT:-N8_sD3hCXul1wHOK1vw:b50vxwq) (no related file):*

Working: We should add (back?) `tags = ["manual"]` to the generated shims.
</blockquote></details>

OK Er, tried it in ca5361d, and still seeing `_{shimmed_binary}.runfiles` present. Punting for now.

___

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/106)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @EricCousineau-TRI, @jwnimmer-tri, and @sloretz)

___
*[`ros2_example_bazel_installed/setup/runfiles_direct_test.sh` line 6 at r3](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8_t9yT-K8vuSITYWU3:-N8_t9yT-K8vuSITYWU4:br8brfl) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/4cac62370a81080b42f64dce3c4f48bec0679fc8/ros2_example_bazel_installed/setup/runfiles_direct_test.sh#L6)):*
> ```Shell
> # This is run directly, not via Bazel, meant to test #105.
> # TODO(eric.cousineau): Remove this pending #107.
> bazel clean --expunge --async
> ```

Working: `--async` and/or `--expunge` seem expensive for CI.

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/106)* status: 0 of 8 files reviewed, all discussions resolved (waiting on @jwnimmer-tri and @sloretz)

___
*[`ros2_example_bazel_installed/setup/runfiles_direct_test.sh` line 6 at r3](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8_t9yT-K8vuSITYWU3:-N8_tPIZDtkf7vy6VvDX:b-896fix) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/4cac62370a81080b42f64dce3c4f48bec0679fc8/ros2_example_bazel_installed/setup/runfiles_direct_test.sh#L6)):*

Working: `--async` and/or `--expunge` seem expensive for CI.
</blockquote></details>

Done.

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/106)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @EricCousineau-TRI, @jwnimmer-tri, and @sloretz)

___
*[`ros2_example_bazel_installed/test/runfiles_test.cc` line 21 at r4](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8_tYl4BLpJjkiI3inu:-N8_tYl5DMi2_c1O7xoo:b2i38v) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/da7583475b183fa8443c36d1932729ef31986d2b/ros2_example_bazel_installed/test/runfiles_test.cc#L21)):*
> ```cpp
> 
>   if (std::getenv("TEST_SRCDIR")) {
>     // I think this is how it works.
> ```

Working: Comments are inaccurate.

EricCousineau-TRI (commented):
CI is looking promising now; +@sloretz for review, please!

> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-:-N8_vPjY-DnAIfMEkO4G:b-ij6c93)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @EricCousineau-TRI and @sloretz)

___
*[`bazel_ros2_rules/ros2/tools/dload_cc.bzl` line 36 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8_bPW5fCEt1zz9EX9q-r2-36:-N8_uc3t9xyJQls-HpF1:bobkfxv) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/b24e7bea7ec2eb285bba75fdb70b71d4cb66fcb8/bazel_ros2_rules/ros2/tools/dload_cc.bzl#L36)):*

Done. Yup yup, I believe that's what I just did here!
</blockquote></details>

OK Er, except in this case, `:bar`'s runfiles are the only things used, at least via itself as entrypoint.
`:foo` runfiles dir could still exist (see above comment about `tags = ["manual"]`), but per your point (a), this only forwards `Runfiles().EnvVars()` for `:bar`.

I believe this should work, even when (hopefully) Bazel stops populating the inner `:foo` via `tags = ["manual"]` (TBH, kinda confused about that).

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/106)* status: 0 of 8 files reviewed, all discussions resolved (waiting on @sloretz)

___
*[`ros2_example_bazel_installed/test/runfiles_test.cc` line 21 at r4](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8_tYl4BLpJjkiI3inu:-N8_v_g9D_H84gRATBcC:b-896fix) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/da7583475b183fa8443c36d1932729ef31986d2b/ros2_example_bazel_installed/test/runfiles_test.cc#L21)):*

Working: Comments are inaccurate.
</blockquote></details>

Done.

jwnimmer-tri:
*[`bazel_ros2_rules/ros2/tools/dload_cc.bzl` line 36 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8_bPW5fCEt1zz9EX9q-r2-36:-N8_wVu-BPpoua86yryr:b-o66z7q) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/b24e7bea7ec2eb285bba75fdb70b71d4cb66fcb8/bazel_ros2_rules/ros2/tools/dload_cc.bzl#L36)):*

OK Er, except in this case, `:bar`'s runfiles are the only things used, at least via itself as entrypoint.
`:foo` runfiles dir could still exist (see above comment about `tags = ["manual"]`), but per your point (a), this only forwards `Runfiles().EnvVars()` for `:bar`.

I believe this should work, even when (hopefully) Bazel stops populating the inner `:foo` via `tags = ["manual"]` (TBH, kinda confused about that).
</blockquote></details>

Yes, bazel might have a lot of files on disk.  For `bazel build :bar && bazel-bin/bar` the only ones we can count on are `bazel-bin/bar.runfiles` which I think you're using here now.

sloretz (approved):
> Reviewed 1 of 5 files at r1, 3 of 7 files at r3, 1 of 1 files at r4, 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/106)* status: all files reviewed, 2 unresolved discussions (waiting on @EricCousineau-TRI)

___
*[`bazel_ros2_rules/ros2/tools/dload_cc.bzl` line 51 at r5](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8aAllM2d6r4XaWXFkT:-N8aAllM2d6r4XaWXFkU:byi4tqk) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/7016f0d0e7923ecf1279dff10e1cb15972d1bee4/bazel_ros2_rules/ros2/tools/dload_cc.bzl#L51)):*
> ```Starlark
>   bool needs_runfiles_env = true;
>   for (const auto& [key, value] : runfiles_env) {{
>     if (nullptr != getenv(key.c_str())) {{
> ```

Is this saying "if anything already set runfiles related environment variables, then use those"? Is the purpose of this to make sure that if `:baz` depends on `:bar` and `baz` launches `:bar`,`:bar` would tell `:foo` to uses `:baz`'s runfiles?
___
*[`bazel_ros2_rules/ros2/tools/dload_cc.bzl` line 59 at r5](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8aBIIuBT8Gubhpc-R3:-N8aBIIuBT8Gubhpc-R4:b-3wpbag) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/7016f0d0e7923ecf1279dff10e1cb15972d1bee4/bazel_ros2_rules/ros2/tools/dload_cc.bzl#L59)):*
> ```Starlark
>     for (const auto& [key, value] : runfiles_env) {{
>       if (setenv(key.c_str(), value.c_str(), 1) != 0) {{
>         std::cerr << "ERROR: failed to set " << key << std::endl;
> ```

I'd recommend prefixing with `DLOAD SHIM ERROR:` to make it easier tell it was sent to stderr by `:bar` and not by `:foo`.

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/106)* status: 7 of 9 files reviewed, all discussions resolved (waiting on @sloretz)

___
*[`bazel_ros2_rules/ros2/tools/dload_cc.bzl` line 51 at r5](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8aAllM2d6r4XaWXFkT:-N8aHBAiEImPObR8-ubd:bbv1s6r) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/7016f0d0e7923ecf1279dff10e1cb15972d1bee4/bazel_ros2_rules/ros2/tools/dload_cc.bzl#L51)):*

Is this saying "if anything already set runfiles related environment variables, then use those"? Is the purpose of this to make sure that if `:baz` depends on `:bar` and `baz` launches `:bar`,`:bar` would tell `:foo` to uses `:baz`'s runfiles?
</blockquote></details>

Done. Didn't put in the example for now.
___
*[`bazel_ros2_rules/ros2/tools/dload_cc.bzl` line 59 at r5](https://reviewable.io/reviews/robotlocomotion/drake-ros/106#-N8aBIIuBT8Gubhpc-R3:-N8aHEOkBVX63gbMJ6-i:b-896fix) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/7016f0d0e7923ecf1279dff10e1cb15972d1bee4/bazel_ros2_rules/ros2/tools/dload_cc.bzl#L59)):*

I'd recommend prefixing with `DLOAD SHIM ERROR:` to make it easier tell it was sent to stderr by `:bar` and not by `:foo`.
</blockquote></details>

Done.

Issue #108, opened 2022-08-04
dload: Clean up items


EricCousineau-TRI:
- [x] Minimize generated code. At present, for C++, the [template code is ~80 lines long](https://github.com/RobotLocomotion/drake-ros/blob/1e4da97bc8ad15a0a595e29c2ea390526dd8b3ed/bazel_ros2_rules/ros2/tools/dload_cc.bzl#L14-L94), with character escaping. For both C++ and Python, we should break this down and provide libraries that are then used via `ros_cc` and `ros_py`.
- [x] Do not re-shim environment variables. At present, there are no checks that we're already shimming. We should ideally say something like `_BAZEL_ROS_RULES_SHIM`. If set, we should not inject *any* new environment variables (that should be controlled by entrypoint, similar to runfiles as shown in #106). If unset, then we set env vars, and set this sentinel. This should happen in both C++ and Python.

\cc @IanTheEngineer @sloretz

sloretz:
The second bullet point is in #110

sloretz:
The first bullet point is in #130

sloretz:
Closing as completed

PR #110, opened 2022-08-09
Skip env modifications if sentinel environment variable set

sloretz:
This is the second half of #108. It makes the C++ and Python shims set the environment variable `_BAZEL_ROS2_RULES_SHIMMED`. If they discover the it was already set then they skip modifying the environment variables. I added a couple `sh_test()`s using `ros_cc_binary()` and `ros_py_binary()` to check that it works by using the knowledge that `AMENT_PREFIX_PATH` is going to be set by the ROS 2 rules.

- - -
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/110)


sloretz (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/110)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @EricCousineau-TRI)

*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N8zwifyBYb6CPC3IBR8:-N8zwifyBYb6CPC3IBR9:b-xvu3k1) (no related file):*
+@EricCousineau-TRI for review please!

___

EricCousineau-TRI (changes requested):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/110)* status: 0 of 6 files reviewed, 7 unresolved discussions (waiting on @EricCousineau-TRI and @sloretz)

*[a discussion](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N8zwifyBYb6CPC3IBR8:-N9-1ahY6gHeO4WJ2D_R:bo647je) (no related file):*

+@EricCousineau-TRI for review please!
</blockquote></details>

Thanks! First pass done, PTAL!

___
___
*[`bazel_ros2_rules/ros2/tools/dload_cc.bzl` line 149 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N9-25YG2vrIR54cSY3g:-N9-25YG2vrIR54cSY3h:b-tgkup8) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/f15874ec9587096b12026dea7855fe450fa2033b/bazel_ros2_rules/ros2/tools/dload_cc.bzl#L149)):*
> ```Starlark
> debuggers like gdb and lldb makes it transparent.
> 
> See do_dload_shim() documentation for further reference.
> ```

Can you document this behavior here?
___
*[`bazel_ros2_rules/ros2/tools/dload_py.bzl` line 24 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N9-2CgLAhYMCt5cBBEC:-N9-2CgLAhYMCt5cBBED:bhavct9) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/f15874ec9587096b12026dea7855fe450fa2033b/bazel_ros2_rules/ros2/tools/dload_py.bzl#L24)):*
> ```Starlark
> 
> 
> SHIMMED_SENTINEL = "_BAZEL_ROS2_RULES_SHIMMED";
> ```

BTW I believe it's 1 newline to separate constants from imports, 2 newlines to separate functions from other things.
___
*[`ros2_example_bazel_installed/BUILD.bazel` line 53 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N9-3V4a-Y3L4YFrJJlb:-N9-3V4a-Y3L4YFrJJlc:bdx5ocu) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/f15874ec9587096b12026dea7855fe450fa2033b/ros2_example_bazel_installed/BUILD.bazel#L53)):*
> ```Starlark
> )
> 
> sh_test(
> ```

At present, this only tests that shimming works as expected at one-level of nesting, but not multiple levels of nesting.

Could you do something like the following?
- Adding command-line argument to both binaries which will `execv` a given absolute path w/ no args; you can use this to see what env vars do when nesting
- In the test program, run 6 total combinations:
  - alone: {cc,py}
  - nested: {cc,py} executing {cc,py} binary
___
*[`ros2_example_bazel_installed/BUILD.bazel` line 53 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N9-3vQd9sW_Ixud1xYs:-N9-3vQd9sW_Ixud1xYt:b-varzdt) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/f15874ec9587096b12026dea7855fe450fa2033b/ros2_example_bazel_installed/BUILD.bazel#L53)):*
> ```Starlark
> )
> 
> sh_test(
> ```

Per below, I think you can consolidate entry point to one `py_test`, and then do combinations inside the Python test.
___
*[`ros2_example_bazel_installed/test/shimmed_sentinel.py` line 6 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N9-33A0EoDFGwzX2KcW:-N9-33A0EoDFGwzX2KcX:b-crizem) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/f15874ec9587096b12026dea7855fe450fa2033b/ros2_example_bazel_installed/test/shimmed_sentinel.py#L6)):*
> ```Python
> assert __name__ == '__main__'
> 
> shimmed = 'yes' if '_BAZEL_ROS2_RULES_SHIMMED' in os.environ else 'no'
> ```

nit For both C++ and Python, consider just using `True/False` as values.

Then you can simply use Python code as-is, and in C++, you can inline value conversion rather than separate `cout <<` statements (for easier-to-read parity).
___
*[`ros2_example_bazel_installed/test/shimmed_sentinel.py` line 9 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N9-2rtsBRIicCwtK3uL:-N9-2rtsBRIicCwtK3uM:b-8zz99z) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/f15874ec9587096b12026dea7855fe450fa2033b/ros2_example_bazel_installed/test/shimmed_sentinel.py#L9)):*
> ```Python
> app_present = 'yes' if 'AMENT_PREFIX_PATH' in os.environ else 'no'
> 
> print(f'shimmed: {shimmed} AMENT_PREFIX_PATH present: {app_present}')
> ```

nit Consider delimiter of newline or comma; at present, single space as `key: value` delimiter is a bit hard to read.
___
*[`ros2_example_bazel_installed/test/shimmed_sentinel_test.sh` line 1 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N9-2iO48T9clTLWN1nw:-N9-2iO48T9clTLWN1nx:b79ncjv) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/f15874ec9587096b12026dea7855fe450fa2033b/ros2_example_bazel_installed/test/shimmed_sentinel_test.sh#L1)):*
> ```Shell
> #!/bin/bash
> ```

nit For more complex testing, it's easier to interpret (and control) Python code.

Can you redux this as Python code?

sloretz (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/110)* status: 0 of 7 files reviewed, 6 unresolved discussions (waiting on @EricCousineau-TRI and @sloretz)

___
*[`bazel_ros2_rules/ros2/tools/dload_py.bzl` line 24 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N9-2CgLAhYMCt5cBBEC:-N92PTYR45fvupbELHyK:bz4biq1) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/f15874ec9587096b12026dea7855fe450fa2033b/bazel_ros2_rules/ros2/tools/dload_py.bzl#L24)):*

BTW I believe it's 1 newline to separate constants from imports, 2 newlines to separate functions from other things.
</blockquote></details>

One newline between import and constant in 37593ca
___
*[`ros2_example_bazel_installed/BUILD.bazel` line 53 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N9-3vQd9sW_Ixud1xYs:-N92W4Es3xZBO5EyofZg:bx7bnl0) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/f15874ec9587096b12026dea7855fe450fa2033b/ros2_example_bazel_installed/BUILD.bazel#L53)):*

Per below, I think you can consolidate entry point to one `py_test`, and then do combinations inside the Python test.
</blockquote></details>

Converted shell script to python in a88585ceb7fdbf96b0796a8fa72d829e47178e17
___
*[`ros2_example_bazel_installed/BUILD.bazel` line 53 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N9-3V4a-Y3L4YFrJJlb:-N92hLGp2LcqcUhq04FZ:b-wd61v1) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/f15874ec9587096b12026dea7855fe450fa2033b/ros2_example_bazel_installed/BUILD.bazel#L53)):*
> At present, this only tests that shimming works as expected at one-level of nesting, but not multiple levels of nesting.

Second level is mocked by setting `_BAZEL_ROS2_RULS_SHIMMED` in the test. I could use `dload_cc_shim()` and `cc_binary()` directly to create a second level of shimming.
___
*[`ros2_example_bazel_installed/test/shimmed_sentinel.py` line 6 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N9-33A0EoDFGwzX2KcW:-N92h6PD6MsRscJS1FJz:b85m9r4) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/f15874ec9587096b12026dea7855fe450fa2033b/ros2_example_bazel_installed/test/shimmed_sentinel.py#L6)):*

nit For both C++ and Python, consider just using `True/False` as values.

Then you can simply use Python code as-is, and in C++, you can inline value conversion rather than separate `cout <<` statements (for easier-to-read parity).
</blockquote></details>

Used inline value conversion in 226aee4
___
*[`ros2_example_bazel_installed/test/shimmed_sentinel.py` line 9 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N9-2rtsBRIicCwtK3uL:-N92hBmjCRhy2IIxvcZo:bchp822) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/f15874ec9587096b12026dea7855fe450fa2033b/ros2_example_bazel_installed/test/shimmed_sentinel.py#L9)):*

nit Consider delimiter of newline or comma; at present, single space as `key: value` delimiter is a bit hard to read.
</blockquote></details>

Outputting json dictionary instead in 226aee4
___
*[`ros2_example_bazel_installed/test/shimmed_sentinel_test.sh` line 1 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N9-2iO48T9clTLWN1nw:-N92W2Zk-mYKKsVeXebY:bx7bnl0) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/f15874ec9587096b12026dea7855fe450fa2033b/ros2_example_bazel_installed/test/shimmed_sentinel_test.sh#L1)):*

nit For more complex testing, it's easier to interpret (and control) Python code.

Can you redux this as Python code?
</blockquote></details>

Converted shell script to python in a88585ceb7fdbf96b0796a8fa72d829e47178e17

sloretz (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/110)* status: 6 of 7 files reviewed, 8 unresolved discussions (waiting on @EricCousineau-TRI)

___
*[`bazel_ros2_rules/ros2/tools/dload_cc.bzl` line 149 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N9-25YG2vrIR54cSY3g:-N93oqfOBKR3B1drUVMQ:bxre4td) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/f15874ec9587096b12026dea7855fe450fa2033b/bazel_ros2_rules/ros2/tools/dload_cc.bzl#L149)):*

Can you document this behavior here?
</blockquote></details>

Documented in c539fc0
___
*[`ros2_example_bazel_installed/test/shimmed_sentinel_test.py` line 1 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N92o3Rd0AaQchuKifRs:-N93p4mW8gOQ6Py1X6Vj:bsjlpao) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/226aee46438a0112ee809ab6e82c50c8a0fc699f/ros2_example_bazel_installed/test/shimmed_sentinel_test.py#L1)):*

nit Shebang is not necessary for Bazel test file.
</blockquote></details>

Removed shebang in 4718316
___
*[`ros2_example_bazel_installed/test/shimmed_sentinel_test.py` line 2 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N92o6gc9KAr7veZmIyu:-N93wrF362TR5j0qd2rV:b-n8lh4) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/226aee46438a0112ee809ab6e82c50c8a0fc699f/ros2_example_bazel_installed/test/shimmed_sentinel_test.py#L2)):*

Can you write docstring of why we are doing this test?
</blockquote></details>

Added module docstring in 6863afa
___
*[`ros2_example_bazel_installed/test/shimmed_sentinel_test.py` line 10 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N92oJfZ1uoPbQgDiNBe:-N93uunv8ZJ3Hequ4lER:b-m17oze) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/226aee46438a0112ee809ab6e82c50c8a0fc699f/ros2_example_bazel_installed/test/shimmed_sentinel_test.py#L10)):*

For clarity, can you require all tests pass `env` explicitly?
</blockquote></details>

Made all callers pass in `env` in 701c798
___
*[`ros2_example_bazel_installed/test/shimmed_sentinel_test.py` line 15 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N92nMTL4hcwONAPp75B:-N93qbIG4nI7IkDjeI9G:bjonmml) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/226aee46438a0112ee809ab6e82c50c8a0fc699f/ros2_example_bazel_installed/test/shimmed_sentinel_test.py#L15)):*

Usage of `r.EnvVars()` is not immediately clear. Can you provide comment as to you're supplying it here?

(I assume b/c you want a completely empty environment, but want to ensure we respect encapsulated `.runfiles`)
</blockquote></details>

I got this from https://github.com/bazelbuild/bazel/blob/21d07586c28b2e569aa4a95d71534123f7780557/tools/python/runfiles/runfiles.py#L48-L58 - I assumed the shimmed executables needed runfiles to find the non-shimmed executable, but I didn't test if that was the case. It seems to work fine without it, so I removed it in b5a242e
___
*[`ros2_example_bazel_installed/test/shimmed_sentinel_test.py` line 24 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N92oFua34FWB2EsyrGF:-N93stfN-C_Vm0q_crIV:b91g83s) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/226aee46438a0112ee809ab6e82c50c8a0fc699f/ros2_example_bazel_installed/test/shimmed_sentinel_test.py#L24)):*

nit There is nothing in this test that quickly asserts that the shim is not present.

Can you add that in `def setUp(self)` here?
</blockquote></details>

Asserts the test is not shimmed in 158037b, though I don't think it would effect the test if it was since the environment of the subprocesses is cleared.
___
*[`ros2_example_bazel_installed/test/shimmed_sentinel_test.py` line 25 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N92onqs-mj6wI6eFPhn:-N93tV4qBRjyPhL-qr7t:beip3zi) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/226aee46438a0112ee809ab6e82c50c8a0fc699f/ros2_example_bazel_installed/test/shimmed_sentinel_test.py#L25)):*

Instead of changing `env` based on same invocation, can you just make two vars in `self.setUp()` and use this directly? e.g.
```
def setUp(self):
  self.runfiles = ...
  self.env_once = self.runfiles.EnvVars()
  self.env_twice = dict(self.env_once)
  self.env_twice[SHIM_KEY] = ""

...
```
</blockquote></details>

I'm not sure I understand. Is 42bcb3a what you had in mind?
___
*[`ros2_example_bazel_installed/test/shimmed_sentinel_test.py` line 35 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N92oDw95P0BPv15JFRc:-N93pqHj7ibko84HtIeq:b-fekdye) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/226aee46438a0112ee809ab6e82c50c8a0fc699f/ros2_example_bazel_installed/test/shimmed_sentinel_test.py#L35)):*

nit Can you use a string constant for the shim here?
</blockquote></details>

Using module scope string constant in 4b070c1

EricCousineau-TRI (changes requested):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/110)* status: 6 of 7 files reviewed, 8 unresolved discussions (waiting on @EricCousineau-TRI and @sloretz)

___
*[`bazel_ros2_rules/ros2/tools/dload_cc.bzl` line 149 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N9-25YG2vrIR54cSY3g:-N940Jt5EnqsKULJ8Ja_:b-yx4kl8) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/f15874ec9587096b12026dea7855fe450fa2033b/bazel_ros2_rules/ros2/tools/dload_cc.bzl#L149)):*

Documented in c539fc0
</blockquote></details>

did you push? I don't see link for commit, nor new revision (r3)

sloretz (commented):
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/110)* status: 6 of 7 files reviewed, 8 unresolved discussions (waiting on @EricCousineau-TRI)

___
*[`bazel_ros2_rules/ros2/tools/dload_cc.bzl` line 149 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N9-25YG2vrIR54cSY3g:-N941nS70HAD5qIzDoFT:b-470t3l) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/f15874ec9587096b12026dea7855fe450fa2033b/bazel_ros2_rules/ros2/tools/dload_cc.bzl#L149)):*

did you push? I don't see link for commit, nor new revision (r3)
</blockquote></details>

I did. I see the commits on github before the comment with my review, but I don't see the new changes on Reviewable either. Maybe they're having technical issues? It has been pretty slow for me today.

EricCousineau-TRI (changes requested):
> Reviewed 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/110)* status: all files reviewed, 4 unresolved discussions (waiting on @sloretz)

___
*[`bazel_ros2_rules/ros2/tools/dload_cc.bzl` line 149 at r1](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N9-25YG2vrIR54cSY3g:-N99F8c2BCFwibI6CQc0:b-fqq45d) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/f15874ec9587096b12026dea7855fe450fa2033b/bazel_ros2_rules/ros2/tools/dload_cc.bzl#L149)):*

I did. I see the commits on github before the comment with my review, but I don't see the new changes on Reviewable either. Maybe they're having technical issues? It has been pretty slow for me today.
</blockquote></details>

OK I see 'em now!
___
*[`bazel_ros2_rules/ros2/tools/dload_cc.bzl` line 149 at r3](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N99FEyg6HflBZYDluNS:-N99FEyg6HflBZYDluNT:bqqfv1f) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/6863afabbacdea6ec9ec774c03af5922aa691e3b/bazel_ros2_rules/ros2/tools/dload_cc.bzl#L149)):*
> ```Starlark
> debuggers like gdb and lldb makes it transparent.
> 
> This shim sets the environment variable `_BAZEL_ROS2_RULES_SHIMMED`. Processes
> ```

While it is good to mention implementation, I think it breaks some encapsulation (e.g. for stability), hence the use of prefixed underscore.

I would recommend just saying something like:
```
This provides a mechanism to only shim once at the top-level; any nested invocations will use the top-level shim environment.
```
___
*[`bazel_ros2_rules/ros2/tools/dload_cc.bzl` line 149 at r3](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N99FWjM0LOvfVpMkLSO:-N99FWjM0LOvfVpMkLSP:bch6un8) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/6863afabbacdea6ec9ec774c03af5922aa691e3b/bazel_ros2_rules/ros2/tools/dload_cc.bzl#L149)):*
> ```Starlark
> debuggers like gdb and lldb makes it transparent.
> 
> This shim sets the environment variable `_BAZEL_ROS2_RULES_SHIMMED`. Processes
> ```

FYI As I was writing the above point, I fear we may run into an issue where binaries come in via `data` deps, but the resolved environment variables may not have sufficient information (e.g. top-level provides env to run A, nested depends on B, B is not a subset of A, thus nested breaks).
This will probably happen sometime in the future.

We may need to interleave things conditionally in the future...

No action necessary, but I want y'all to be aware.
\cc @iantheengineer 
___
*[`ros2_example_bazel_installed/test/shimmed_sentinel_test.py` line 15 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N92nMTL4hcwONAPp75B:-N99GCaH7fNXXVXEhkuZ:b-3i4ph) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/226aee46438a0112ee809ab6e82c50c8a0fc699f/ros2_example_bazel_installed/test/shimmed_sentinel_test.py#L15)):*

I got this from https://github.com/bazelbuild/bazel/blob/21d07586c28b2e569aa4a95d71534123f7780557/tools/python/runfiles/runfiles.py#L48-L58 - I assumed the shimmed executables needed runfiles to find the non-shimmed executable, but I didn't test if that was the case. It seems to work fine without it, so I removed it in b5a242e
</blockquote></details>

Ah... my above was only defecting lack of documentation, not the usage.

I believe this may now be defective, in the same vein as in https://github.com/RobotLocomotion/drake-ros/pull/106#pullrequestreview-1061069826 (breaking out of sandbox when referring to runfiles)
Can you double check with `bazel clean`?
___
*[`ros2_example_bazel_installed/test/shimmed_sentinel_test.py` line 24 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N92oFua34FWB2EsyrGF:-N99GlZe0jyjeSb0WzDN:b-npi1j5) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/226aee46438a0112ee809ab6e82c50c8a0fc699f/ros2_example_bazel_installed/test/shimmed_sentinel_test.py#L24)):*

Asserts the test is not shimmed in 158037b, though I don't think it would effect the test if it was since the environment of the subprocesses is cleared.
</blockquote></details>

OK Ah, derp... Good point! Wel, yeah, feel free to keep or remove this.
___
*[`ros2_example_bazel_installed/test/shimmed_sentinel_test.py` line 25 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N92onqs-mj6wI6eFPhn:-N99Gr6T9gS6hwQH3guz:bobp2ay) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/226aee46438a0112ee809ab6e82c50c8a0fc699f/ros2_example_bazel_installed/test/shimmed_sentinel_test.py#L25)):*

I'm not sure I understand. Is 42bcb3a what you had in mind?
</blockquote></details>

Close, but see above concern about cross-runfile pollution when removing `r.EnvVars()`.
___
*[`ros2_example_bazel_installed/test/shimmed_sentinel_test.py` line 24 at r3](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N99GfRS0T4ng2u8Sicc:-N99GfRS0T4ng2u8Sicd:b-e5uwm7) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/6863afabbacdea6ec9ec774c03af5922aa691e3b/ros2_example_bazel_installed/test/shimmed_sentinel_test.py#L24)):*
> ```Python
>     """Run a bazel executable target and return stdout."""
>     r = runfiles.Create()
>     if env is None:
> ```

I believe this is now dead code. Can you remove this `if` statement?

sloretz (commented):
> Reviewed 5 of 6 files at r2, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/110)* status: all files reviewed, 4 unresolved discussions (waiting on @EricCousineau-TRI)

___
*[`bazel_ros2_rules/ros2/tools/dload_cc.bzl` line 149 at r3](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N99FEyg6HflBZYDluNS:-NAGt2a21FIP86zIVaTT:b-eiahyn) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/6863afabbacdea6ec9ec774c03af5922aa691e3b/bazel_ros2_rules/ros2/tools/dload_cc.bzl#L149)):*

While it is good to mention implementation, I think it breaks some encapsulation (e.g. for stability), hence the use of prefixed underscore.

I would recommend just saying something like:
```
This provides a mechanism to only shim once at the top-level; any nested invocations will use the top-level shim environment.
```
</blockquote></details>

Wrote less detail in ef338f3
___
*[`ros2_example_bazel_installed/test/shimmed_sentinel_test.py` line 15 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N92nMTL4hcwONAPp75B:-NAGyc_d1PYOwLiI5dmZ:b-u2vpsh) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/226aee46438a0112ee809ab6e82c50c8a0fc699f/ros2_example_bazel_installed/test/shimmed_sentinel_test.py#L15)):*
> I believe this may now be defective, in the same vein as in #106 (review) (breaking out of sandbox when referring to runfiles)
> Can you double check with bazel clean?

Ah, meaning the runfiles for `shimmed_sentinel_cc` and `shimmed_sentinel_py` don't exist if only `bazel test //:shimmed_sentinel_test` is run, so they have to be given the test's runfiles to work. I double checked that was the issue, restored using `r.EnvVars()`, and documented it in f2b7c09
___
*[`ros2_example_bazel_installed/test/shimmed_sentinel_test.py` line 25 at r2](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N92onqs-mj6wI6eFPhn:-NAH8w_h7kS1S3628q9w:b-vo4cya) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/226aee46438a0112ee809ab6e82c50c8a0fc699f/ros2_example_bazel_installed/test/shimmed_sentinel_test.py#L25)):*

Close, but see above concern about cross-runfile pollution when removing `r.EnvVars()`.
</blockquote></details>

Added back `r.EnvVars()`
___
*[`ros2_example_bazel_installed/test/shimmed_sentinel_test.py` line 24 at r3](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-N99GfRS0T4ng2u8Sicc:-NAGtNJy6gvbOFvvDSoK:b-cr30mg) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/6863afabbacdea6ec9ec774c03af5922aa691e3b/ros2_example_bazel_installed/test/shimmed_sentinel_test.py#L24)):*

I believe this is now dead code. Can you remove this `if` statement?
</blockquote></details>

Removed in 5557220

EricCousineau-TRI (approved):
:lgtm: with nit on comments

> Reviewed 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-:-NAKn9sm-a_Rkq8v9Tao:b-6n47h9)* status: all files reviewed, 1 unresolved discussion (waiting on @sloretz)

___
*[`ros2_example_bazel_installed/test/shimmed_sentinel_test.py` line 29 at r4](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-NAKmdl1BnHD3BSHo1_I:-NAKmdl1BnHD3BSHo1_J:b-g5v2s1) ([raw file](https://github.com/robotlocomotion/drake-ros/blob/f2b7c092141e7467d6abb7e134a57a1fe2e805db/ros2_example_bazel_installed/test/shimmed_sentinel_test.py#L29)):*
> ```Python
>     # if bazel was explicitly asked to build them, meaning this test would fail
>     # if it were the only thing run from a clean build. This gives the binaries
>     # access to the test's runfiles, which should have everything they need.
> ```

nit While this description is accurate, the side effect ("won't exist unless") seems to indicate a workaround, but IMO it is an invalid workaround because it breaks encapsulation (in terms of build commands, hashing for the target, etc.)

Consider minimizing this text, and reference something like `drake-ros#106` or what not, e.g.:
```
# Ensure we use this target's runfiles for data encapsulation (#106).
```

sloretz:
@EricCousineau-TRI , I'm not sure what's going on, but I can't comment on Reviewable. After I type about 15 characters reviewable resets and clears the text box I'm typing in. This is on firefox and the only extension I have is LastPass.

I reworded the comment to not imply a workaround and referenced #106 in 4a51e04. I didn't shorten it too much because I'd need a longer description to understand why runfiles were being used.

> IMO it is an invalid workaround because it breaks encapsulation (in terms of build commands, hashing for the target, etc.)

Couldn't we also say bazel is requiring us to break encapsulation by not building target specific runfiles? If I understand correctly, giving each binary the test's runfiles is giving them access to the runfiles of all the other binaries too.

EricCousineau-TRI (approved):
hm... not sure if I know what is going on for Reviewable. I am not having issues, and I am using Firefox w/ many plugins, including LastPass.
Maybe try clearing cache + cookies?

> Couldn't we also say bazel is requiring us to break encapsulation by not building target specific runfiles?

No, I don't think so.
I think Bazel's contract is that it will provide `.runfiles` for only the target you build, and that caching is probably only dependent on the `.runfiles` therein.
That seems reasonable to me. Asking to cross over into separate `.runfiles` seems like it could break that contract.

> Reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/110#-:-NALmrNk18rWl8SPOcwS:b-at9gb3)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sloretz)

Issue #125, opened 2022-08-26
cyclone dds: What does `/ParticipantIndex="none"` mean for usage in ROS2?


EricCousineau-TRI:
Wanted to check on rationale `ROS_DOMAIN_ID`, confirmed that yeah it's just direct link to DDS Domain ID, and it is used to partition port numbers.

Looked at docs for limit on number of participants:
https://docs.ros.org/en/humble/Concepts/About-Domain-ID.html#participant-constraints

Then also saw the Eclipse DDS stuff that @hidmic referenced to in `rmw_isolation`:
https://cyclonedds.io/docs/cyclonedds/0.9.1/config.html#controlling-port-numbers
https://github.com/eclipse-cyclonedds/cyclonedds/blob/0.9.1/src/core/ddsi/src/ddsi_portmapping.c#L39-L44
https://github.com/eclipse-cyclonedds/cyclonedds/blob/0.9.1/src/core/ddsi/include/dds/ddsi/ddsi_cfgelems.h#L1883-L1902
Looked at usage of `ParticipantIndex="none"` here too:
https://github.com/RobotLocomotion/drake-ros/blob/c7f9c362c5b4f2071235e61bc64f8ac1ad0c42f5/bazel_ros2_rules/ros2/resources/rmw_isolation/rmw_isolation.py#L153-L162

Questions:

1. I am a bit confused about when multicast or unicast is used. Are there ROS-specific docs about this? (may relate #103)
2. Do we need to set `ParticipantGain` at all if can use auto-resolve its port (and thus overcome limitations on participant IDs)? (from code linked above, it appears not to be)
3. Should this be mentioned to some extent in ROS documentation?

\cc @sloretz @cottsay

sloretz:
> I am a bit confused about when multicast or unicast is used. Are there ROS-specific docs about this?
> [...]
> Should this be mentioned to some extent in ROS documentation?

I don't know of any ROS specific documentation about when multicast or unicast is used, but it would be specific to a single RMW implementation. CycloneDDS has some documentation about when multicast is used here: https://cyclonedds.io/docs/cyclonedds/latest/config/cyclonedds_specifics.html#multicasting . The information is probably in the [DDSI-RTPS spec](https://www.omg.org/spec/DDSI-RTPS/2.5/PDF), but it would take a while to find it.

> Do we need to set `ParticipantGain` at all if can use auto-resolve its port (and thus overcome limitations on participant IDs)? (from code linked above, it appears not to be)

[...]

PR #126, opened 2022-08-26
rmw_isolation: Disable need for participant gain for cyclone dds

EricCousineau-TRI:
Should not be necessary given ParticipantIndex="none"

See #125

- - -
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/126)


EricCousineau-TRI:
(adding this to the queue, but can defer for the next couple of weeks or so)

EricCousineau-TRI:
whelp, tests fail (kind of a good thing... maybe?), but I guess that means I don't understand the docs from how I've skimmed it.

PR #210, opened 2023-01-11
Update for Rolling PublisherBase and SubscriptionBase API change

sloretz:
APIs we're using changed in https://github.com/ros2/rclcpp/pull/2066 which is released in RCLCPP 18.0.0 in ROS ROlling. This PR uses the updated API.

Fixes #208

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/210)

PR #216, opened 2023-02-02
Switch everything over to Jammy

sloretz:
Follow up to #206 
Requires #210

This switches all of the bazel builds to Jammy

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/216)


cottsay (commented):
If you grep through this repo, there is mention in the README files of Focal. We should change those references to Humble now that we're no longer testing on Focal at all.

EricCousineau-TRI:
Done. Fixed references and rebased on latest `main`.

EricCousineau-TRI:
Failures as expected, going to go ahead and merge forward for now!

Issue #115, opened 2022-08-10
ros2_example_bazel_installed: Should have simple cc,py example of using RMW isolation


EricCousineau-TRI:
Per Anzu 9275, we should have a simple cc and py test that exercises RMW isolation.
e.g. `rmw_isolation_example_cc_test`, `rmw_isolation_example_py_test`

Both can simply publish a float, and do no other verification (for now).

They should then be referenced by:
https://github.com/RobotLocomotion/drake-ros/tree/eba927fd3ecf0ce2819200e0a8176e7d3f366e2a/bazel_ros2_rules/ros2#tools

What our extended usage looks like in Anzu:
https://gist.github.com/EricCousineau-TRI/bf120a42d7aa835037642789a4c3c985

\cc @IanTheEngineer

adityapande-1995:
I see a test here for isolated listener and talker in both C++ and python : https://github.com/RobotLocomotion/drake-ros/blob/main/bazel_ros2_rules/ros2/resources/rmw_isolation/test/isolated_listener.py. To we require something more to resolve this issue ? cc @sloretz

adityapande-1995:
Also a duplicate issue probably : https://github.com/RobotLocomotion/drake-ros/issues/77

sloretz:
> Do we require something more to resolve this issue ? cc @sloretz

I think I recall this issue was to have an example test in the `ros2_example_bazel_installed` folder.

That folder is a separate bazel workspace, which matches what our downstream users would use. The C++/Python files will probably look similar, so IIUC the value in putting tests there is in showing what bazel rules to `load()` and run in the `WORKSPACE` and `BUILD.bazel` files.

EricCousineau-TRI:
Yes, a very clear, user-based example of both C++ and Python of how to use isolation is my intent.

Having it in a more focused example workspace makes sense too! note that existing rmw_isolation tests are more for unittesting, less for exemplars.

PR #232, opened 2023-02-22
RMW example tests in python and C++

adityapande-1995:
Resolves https://github.com/RobotLocomotion/drake-ros/issues/115, by adding standalone cc and py tests.

Spawns multiple talkers and listeners isolated by rmw configurations, so that there is no cross talk between them.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232)


adityapande-1995:
I just merged the ``main`` from this repo into my fork's branch, this should fix CI

EricCousineau-TRI (changes requested):
> Reviewed 2 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @adityapande-1995)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NP-xLdHBG7TeSAAKrLR:-NP-xLdHBG7TeSAAKrLS:beppyeh) (no related file):*
Ah, sorry for not better clarifying!

We already having decent testing of the isolation mechanism; however, what we don't have (the intent of #115) was a good, clear *example* of showing the usage of isolation in both C++ and Python tests.
e.g. if a user asks, "how do I isolate traffic? what's a good example", I'd like to point them here.

Can you re-cater this PR to that intent? e.g.
- Ensure both independently use isolation, and add overview comments stating that this is the objective of each of the source files.
- Make the ROS 2 usage as simple as possible for both C++ and Python. e.g. only have one node that publishes and then exits very quickly - e.g. 2-3 lines of code. Just add a comment that we can publish and should see no traffic on our primary system.

Does this make sense?

___

adityapande-1995:

Ah, sorry for not better clarifying!

We already having decent testing of the isolation mechanism; however, what we don't have (the intent of #115) was a good, clear *example* of showing the usage of isolation in both C++ and Python tests.
e.g. if a user asks, "how do I isolate traffic? what's a good example", I'd like to point them here.

Can you re-cater this PR to that intent? e.g.
- Ensure both independently use isolation, and add overview comments stating that this is the objective of each of the source files.
- Make the ROS 2 usage as simple as possible for both C++ and Python. e.g. only have one node that publishes and then exits very quickly - e.g. 2-3 lines of code. Just add a comment that we can publish and should see no traffic on our primary system.

Does this make sense?
</blockquote></details>

Hello ! If I understand correctly, this would mean more like a tutorial for the end user ? I can imagine a README file, with instructions on how to run the nodes. We can have 2 directories, one for the cpp talker and one for the python talker, each publishing the same float on the same topic, but isolated by rmw. The talkers would be minimal, and just publish a float a couple of times and then exit.

Both of them would be built by ros_cc/py_test() rules. Does that sound okay ?

adityapande-1995:
<details><summary><i>Previously, adityapande-1995 (Aditya Pande) wrote…</i></summary><blockquote>

Hello ! If I understand correctly, this would mean more like a tutorial for the end user ? I can imagine a README file, with instructions on how to run the nodes. We can have 2 directories, one for the cpp talker and one for the python talker, each publishing the same float on the same topic, but isolated by rmw. The talkers would be minimal, and just publish a float a couple of times and then exit.

Both of them would be built by ros_cc/py_test() rules. Does that sound okay ?
</blockquote></details>

On a slight tangent, @sloretz and I were discussing offline, and it seems the original bash script to test rmw isolation isn't running the talker-listener pairs in parallel, and the tests here are. Would it be worth moving these cc/py test scripts to bazel_ros2_rules ?

EricCousineau-TRI (changes requested):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @sloretz)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NP-xLdHBG7TeSAAKrLR:-NP0BjbcBKuFY-3fTuuh:b-7lwd6j) (no related file):*
> [...] it seems the original bash script to test rmw isolation isn't running the talker-listener pairs in parallel, and the tests here are. Would it be worth moving these cc/py test scripts to bazel_ros2_rules ?

Ooh, yup! Can I ask you how you observed this? (Any chance I can easily reproduce your observation?)

> [...] this would mean more like a tutorial for the end user ? I can imagine a README file, with [...]

Hm... I think a README could be helpful for pointing to the targets to look at it, but something this technical I'd prefer an "easy-to-verify" example (that is, real code that is built and tested by CI).

___

adityapande-1995:

> [...] it seems the original bash script to test rmw isolation isn't running the talker-listener pairs in parallel, and the tests here are. Would it be worth moving these cc/py test scripts to bazel_ros2_rules ?

Ooh, yup! Can I ask you how you observed this? (Any chance I can easily reproduce your observation?)

> [...] this would mean more like a tutorial for the end user ? I can imagine a README file, with [...]

Hm... I think a README could be helpful for pointing to the targets to look at it, but something this technical I'd prefer an "easy-to-verify" example (that is, real code that is built and tested by CI).
</blockquote></details>

About the bash script, I haven't tested this explicitly, but I suspect the line here : https://github.com/RobotLocomotion/drake-ros/blob/bc34b5c59dcafb6040f2896d88536abe7d5f9192/bazel_ros2_rules/ros2/resources/rmw_isolation/test/rmw_isolation_test.sh#L50, doesn't have a "&" at the end, and so is triggering the ``test_isolation`` sequentially. It spawns a talker, then a listener, and then shuts down both of them. I suspect the talkers and not running at the same time, which would test for cross-talk.

For this ticket, I was thinking of having 2 directories, one with code for the isolated cpp talker, and the other for isolated python talker. The readme would explain how the isolation works, and how to compile and run the talkers. There will be a ``ros2_py/cc_test()`` rule to build and test these for CI. How does that sound ?

adityapande-1995:
Moved the tests here to replace the bash test : https://github.com/RobotLocomotion/drake-ros/pull/256

EricCousineau-TRI (changes requested):
:lgtm: with two blocking items on import and when/where to isolate

> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-:-NSrdw-x44vh1DZJtPsm:b-p2swhn)* status: all files reviewed, 11 unresolved discussions (waiting on @adityapande-1995)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NP-xLdHBG7TeSAAKrLR:-NSrdtbfBYcCkBTL3ssO:bj36fus) (no related file):*
<details><summary><i>Previously, adityapande-1995 (Aditya Pande) wrote…</i></summary><blockquote>

About the bash script, I haven't tested this explicitly, but I suspect the line here : https://github.com/RobotLocomotion/drake-ros/blob/bc34b5c59dcafb6040f2896d88536abe7d5f9192/bazel_ros2_rules/ros2/resources/rmw_isolation/test/rmw_isolation_test.sh#L50, doesn't have a "&" at the end, and so is triggering the ``test_isolation`` sequentially. It spawns a talker, then a listener, and then shuts down both of them. I suspect the talkers and not running at the same time, which would test for cross-talk.

For this ticket, I was thinking of having 2 directories, one with code for the isolated cpp talker, and the other for isolated python talker. The readme would explain how the isolation works, and how to compile and run the talkers. There will be a ``ros2_py/cc_test()`` rule to build and test these for CI. How does that sound ?
</blockquote></details>

OK Thanks! Yeah, this setup looks great.

___
___
*[`ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.cc` line 59 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NSrdJR1EPprbGgyLm7Q:-NSrdJR20tUzQUFI-l3M:b85mqdl) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/fd4a07b15950af66573659cc0c33d2399a9443ed/ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.cc#L59)):*
> ```cpp
> 
>   rclcpp::Subscription<std_msgs::msg::Float64>::SharedPtr subscription_;
>   size_t expectedMessagesCount_{0u};
> ```

nit This is not read?

Is there a way to briefly test this? e.g. once message count exceeds 2, shutdown and exit listiner?
___
*[`ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.cc` line 65 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NSrcZQZ2Cf1FNO_-2tR:-NSrcZQ_CdvUHjnIkzeh:b-1ung64) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/fd4a07b15950af66573659cc0c33d2399a9443ed/ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.cc#L65)):*
> ```cpp
> pid_t LaunchNode(int argc, char* argv[], const std::string& directory_path,
>     const std::string& node_type="talker"){
>   // Launch a new process.
> ```

nit It may be unclear to people if they should prefer this over executing a separate program.

```
// Note that you can also execute a separate program using functions like `system()` if you so choose.
```
___
*[`ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.cc` line 98 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NSrasAuBxay7JBaxyHH:-NSrasAuBxay7JBaxyHI:b-agp1rw) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/fd4a07b15950af66573659cc0c33d2399a9443ed/ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.cc#L98)):*
> ```cpp
>   }
> 
>   // The talker and listener processes are supplied a common directory_path
> ```

Rather than isolate in each separate fork, can you instead isolate at the top-level (at this spot)?

This is an easy way to have coordinated yet isolated subset of network traffic, and no need to pass arguments to other subprocesses that may get launched in other ways.
You should then be able to remove the need for `directory_path`.
___
*[`ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.cc` line 104 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NSrd7NdDgQWZTPOBrlf:-NSrd7NdDgQWZTPOBrlg:bccepfp) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/fd4a07b15950af66573659cc0c33d2399a9443ed/ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.cc#L104)):*
> ```cpp
>   auto talker_process = LaunchNode(argc, argv, directory_path, "talker");
>   // Wait for the talker to start.
>   sleep(1.0);
> ```

nit This is a very long wait.

Rather than sleepeing, can we instead just wait for listener to exit normally, then terminate?
We can rely on bazel test timeout to kill if things aren't rigged up correctly.
___
*[`ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.cc` line 112 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NSrc8sv8yXYIXdTLgPB:-NSrc8sv8yXYIXdTLgPC:b-jlgwm0) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/fd4a07b15950af66573659cc0c33d2399a9443ed/ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.cc#L112)):*
> ```cpp
> 
>   // Kill the nodes.
>   kill(talker_process, SIGTERM);
> ```

nit Why not `SIGINT`? I think we should prefer slightly less aggressive shutdown if/when possible.
___
*[`ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py` line 46 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NSrbTIf9srkTkSfY7jn:-NSrbTIf9srkTkSfY7jo:bzgi0wd) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/fd4a07b15950af66573659cc0c33d2399a9443ed/ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py#L46)):*
> ```Python
> # Launch a process for the talker or the listener.
> def launch_node(directory_path, node_type="talker"):
>     # Create a logging directory for ROS.
> ```

nit Can you add TODO for me to use something like `make_unique_ros_isolation_env` inside of `bazel_ros_testing.py` to replace some of this?
___
*[`ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py` line 50 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NSrbiDF1p8nGdtqC8L5:-NSrbiDG1jQ-DIUU2HLE:b-h08ndq) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/fd4a07b15950af66573659cc0c33d2399a9443ed/ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py#L50)):*
> ```Python
>     os.environ["ROS_LOG_DIR"] = log_directory.name
> 
>     # RMW isolation is implemented using a dedicated path
> ```

I don't believe this comment is correct.

Any resolution of configuration is done at time of the function call, not import.
Can you remove the comment, and hoist import to the top?
___
*[`ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py` line 53 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NSrbbnT47-2RI34vd98:-NSrbbnT47-2RI34vd99:b-blvo94) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/fd4a07b15950af66573659cc0c33d2399a9443ed/ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py#L53)):*
> ```Python
>     # for the config file. Invoking isolate_rmw_by_path()
>     # before rclpy.init() isolates this process and the nodes in it.
>     from rmw_isolation import isolate_rmw_by_path
> ```

nit Why import here?
___
*[`ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py` line 76 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NSrb1xc8gM8IW56xGBa:-NSrb1xc8gM8IW56xGBb:b6xfk0d) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/fd4a07b15950af66573659cc0c33d2399a9443ed/ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py#L76)):*
> ```Python
>     directory_path = os.path.join(os.getcwd(), "test_node_pair")
>     if not os.path.exists(directory_path): os.mkdir(directory_path)
> ```

Same as above - can you isolate at the top level before launching / forking new subprocs?
___
*[`ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py` line 81 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NSrcJ108GOytan09vNG:-NSrcJ11BL7s2GZyDMhc:b-tsivtq) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/fd4a07b15950af66573659cc0c33d2399a9443ed/ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py#L81)):*
> ```Python
>     # from rest of the system. For e.g, if one were to run a new subscriber on the /chatter topic,
>     # the data published by the talker would not be visible.
>     talker_process = Process(target=launch_node, args=(directory_path, "talker"))
> ```

nit It may be unclear to people if they should prefer this over something like `subprocess`.

```
# Note that you can also use something like subprocess.Popen() if that is more convenient for your workflow.
```
___
*[`ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py` line 86 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NSrdl3L4r5YnYanEDi8:-NSrdl3M9yCxv7RuQxzt:bf1ci2d) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/fd4a07b15950af66573659cc0c33d2399a9443ed/ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py#L86)):*
> ```Python
>     # Wait for the talker to start.
>     talker_process.start()
>     sleep(1.0)
> ```

nit Same as above - can we instead let the listener shutdown / exit once it's received two messages, then stop testing after that?

adityapande-1995:
*[`ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.cc` line 59 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NSrdJR1EPprbGgyLm7Q:-NT0uCeD9pRLLUFunou7:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/fd4a07b15950af66573659cc0c33d2399a9443ed/ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.cc#L59)):*

nit This is not read?

Is there a way to briefly test this? e.g. once message count exceeds 2, shutdown and exit listiner?
</blockquote></details>

Done

adityapande-1995:
*[`ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.cc` line 65 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NSrcZQZ2Cf1FNO_-2tR:-NT0uELl6zvs1AlTlG2-:bkmcjp5) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/fd4a07b15950af66573659cc0c33d2399a9443ed/ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.cc#L65)):*

nit It may be unclear to people if they should prefer this over executing a separate program.

```
// Note that you can also execute a separate program using functions like `system()` if you so choose.
```
</blockquote></details>

Done, added the comment.

adityapande-1995:
*[`ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.cc` line 98 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NSrasAuBxay7JBaxyHH:-NT0uHbh2ZpuAk-NaYk-:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/fd4a07b15950af66573659cc0c33d2399a9443ed/ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.cc#L98)):*

Rather than isolate in each separate fork, can you instead isolate at the top-level (at this spot)?

This is an easy way to have coordinated yet isolated subset of network traffic, and no need to pass arguments to other subprocesses that may get launched in other ways.
You should then be able to remove the need for `directory_path`.
</blockquote></details>

Done

adityapande-1995:
*[`ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.cc` line 104 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NSrd7NdDgQWZTPOBrlf:-NT0uJbI6Y5X2pGldd3u:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/fd4a07b15950af66573659cc0c33d2399a9443ed/ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.cc#L104)):*

nit This is a very long wait.

Rather than sleepeing, can we instead just wait for listener to exit normally, then terminate?
We can rely on bazel test timeout to kill if things aren't rigged up correctly.
</blockquote></details>

Done

adityapande-1995:
*[`ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.cc` line 112 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NSrc8sv8yXYIXdTLgPB:-NT0uL4S4SkIq_-KB0wI:bjp0s4y) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/fd4a07b15950af66573659cc0c33d2399a9443ed/ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.cc#L112)):*

nit Why not `SIGINT`? I think we should prefer slightly less aggressive shutdown if/when possible.
</blockquote></details>

Change to SIGINT, done

adityapande-1995:
*[`ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py` line 46 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NSrbTIf9srkTkSfY7jn:-NT0uOmS6eIlbU6Tngv1:bb84u71) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/fd4a07b15950af66573659cc0c33d2399a9443ed/ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py#L46)):*

nit Can you add TODO for me to use something like `make_unique_ros_isolation_env` inside of `bazel_ros_testing.py` to replace some of this?
</blockquote></details>

Turns out that was not required, removed.

adityapande-1995:
*[`ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py` line 50 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NSrbiDF1p8nGdtqC8L5:-NT0uTIB8Et_s2wfhuYQ:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/fd4a07b15950af66573659cc0c33d2399a9443ed/ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py#L50)):*

I don't believe this comment is correct.

Any resolution of configuration is done at time of the function call, not import.
Can you remove the comment, and hoist import to the top?
</blockquote></details>

Done

adityapande-1995:
*[`ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py` line 53 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NSrbbnT47-2RI34vd98:-NT0uUy99cfsGNmUbXvl:b-7129z9) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/fd4a07b15950af66573659cc0c33d2399a9443ed/ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py#L53)):*

nit Why import here?
</blockquote></details>

Moved the import to the top

adityapande-1995:
*[`ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py` line 76 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NSrb1xc8gM8IW56xGBa:-NT0uYGy3r-AHeXJcYsc:b-896fjb) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/fd4a07b15950af66573659cc0c33d2399a9443ed/ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py#L76)):*

Same as above - can you isolate at the top level before launching / forking new subprocs?
</blockquote></details>

Done

adityapande-1995:
*[`ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py` line 81 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NSrcJ108GOytan09vNG:-NT0uZgbB2sYYoL_geq8:b-6ecklb) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/fd4a07b15950af66573659cc0c33d2399a9443ed/ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py#L81)):*

nit It may be unclear to people if they should prefer this over something like `subprocess`.

```
# Note that you can also use something like subprocess.Popen() if that is more convenient for your workflow.
```
</blockquote></details>

Added the comment, done

adityapande-1995:
*[`ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py` line 86 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NSrdl3L4r5YnYanEDi8:-NT0uaTVArthVFc5DHZ2:b-ab5v1l) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/fd4a07b15950af66573659cc0c33d2399a9443ed/ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py#L86)):*

nit Same as above - can we instead let the listener shutdown / exit once it's received two messages, then stop testing after that?
</blockquote></details>

Removed sleep(), subscriber now exits after 2 msgs are received. Done

EricCousineau-TRI (commented):
Looks like there is a `rolling` failure, but it smells like an API change.
Is it possible to address that in a separate PR?

> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-:-NT0zS1H97wART1VKn6q:b-hva6ug)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @adityapande-1995)

___
*[`ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.cc` line 55 at r3](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NT0zEZJ74lR0RsBxluz:-NT0zEZK6MDhf2YEKWav:b-bs9jy6) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/9df47ed3115e71b8e3cac05f60cefc1a75537cd9/ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.cc#L55)):*
> ```cpp
> 
>  private:
>   void topicCallback(const std_msgs::msg::Float64::SharedPtr msg) {
> ```

nit (here and elsewhere) Should use `CamelCase` for methods, `snake_case` for O(1) methods and variables.
___
*[`ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.cc` line 56 at r3](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NT0zCcdEwU_CDE96GQo:-NT0zCcdEwU_CDE96GQp:b-pdh13r) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/9df47ed3115e71b8e3cac05f60cefc1a75537cd9/ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.cc#L56)):*
> ```cpp
>  private:
>   void topicCallback(const std_msgs::msg::Float64::SharedPtr msg) {
>     ++expectedMessagesCount_;
> ```

nit Name `expected` seems wrong here. Perhaps remove?

EricCousineau-TRI (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232)* status: all files reviewed (commit messages unreviewed), 3 unresolved discussions (waiting on @adityapande-1995)

___
*[`ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py` line 43 at r3](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NT0z_I69Emc5GMy-e4F:-NT0z_I69Emc5GMy-e4G:b-ol26xh) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/9df47ed3115e71b8e3cac05f60cefc1a75537cd9/ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py#L43)):*
> ```Python
> 
>     def _topic_callback(self, _):
>         self._expected_messages_received += 1
> ```

nit `expected` here seems wrong. Can you remove that prefix?

adityapande-1995:
*[`ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py` line 43 at r3](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-NT0z_I69Emc5GMy-e4F:-NTFVyOJ4ljKQS3RAAdj:b-896fjb) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/9df47ed3115e71b8e3cac05f60cefc1a75537cd9/ros2_example_bazel_installed/ros2_example_apps/test/rmw_isolation_example.py#L43)):*

nit `expected` here seems wrong. Can you remove that prefix?
</blockquote></details>

Done

adityapande-1995:
``fastcdr`` also seems to be causing some issue. Probably some package below drake-ros is not able to find it.

EricCousineau-TRI (approved):
Aye, gotcha - I can file issue if  y'all haven't already. Did y'all already discuss w/ Calder + Ian on approach to handle?

> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/232#-:-NTFWSaNAwRpXJeFU5tH:bko6bux)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @adityapande-1995)

adityapande-1995:
Ah, not yet. I can spend some time looking at it, as this will affect CI in all drake-ros PRs. cc @IanTheEngineer

EricCousineau-TRI:
Aye, filed #274

adityapande-1995:
I'll rebase this on the new ``main``, CI should be good to go now.

EricCousineau-TRI:
Sweet! But did you push? I still see CI results from last week still.

EricCousineau-TRI:
Sweet, thanks!

PR #233, opened 2023-02-25
[bazel_ros2_rules] Don't re-exec in ros_cc_binary and ros_cc_test

jwnimmer-tri:
We can directly replace the user's `main()`. This simplifies downstream build logic by preserving the invariant that a binary is one program, not two.

---

For a genrule in Anzu, we need to execute a `ros_cc_binary` but we don't have access to its runfiles. The only way that can work is when it has no runfiles, so therefore it can't have a shim.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233)


jwnimmer-tri:
*[`bazel_ros2_rules/ros2/tools/dload_cc.bzl` line 14 at r1](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NP92TD_8xNOGm0GUG_x:-NP92TD_8xNOGm0GUG_y:bu0zkox) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/b3de699db70791b35c836dba39f871cd935081de/bazel_ros2_rules/ros2/tools/dload_cc.bzl#L14)):*
> ```Starlark
> extern "C" int __real_main(int argc, char** argv);
> 
> extern "C" int __wrap_main(int argc, char** argv) {{
> ```

Working

This doesn't handle non-conforming programs that try to use a 3-arg main.

jwnimmer-tri (commented):
+@EricCousineau-TRI for review or delegation, please.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-:-NP9H_GJ6cJDILic7S6H:b3610v7)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @EricCousineau-TRI and @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NP9HQ6j2IeLu92z33T7:-NP9HQ6j2IeLu92z33T8:bnz3pjo) (no related file):*
Working

Do not merge until Anzu PR 10149 merges first.

___
___
*[`bazel_ros2_rules/ros2/tools/dload_cc.bzl` line 14 at r1](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NP92TD_8xNOGm0GUG_x:-NP9HWd50dzpv4OiZ3HE:bd0totr) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/b3de699db70791b35c836dba39f871cd935081de/bazel_ros2_rules/ros2/tools/dload_cc.bzl#L14)):*
<details><summary><i>Previously, jwnimmer-tri (Jeremy Nimmer) wrote…</i></summary><blockquote>

Working

This doesn't handle non-conforming programs that try to use a 3-arg main.
</blockquote></details>

Done (via documentation in the macros).

EricCousineau-TRI (changes requested):
:lgtm: at a broad level

I'm a bit scared of advanced linker techniques as it's more than what I've wanted to learn.
However, if this passes both local and Anzu CI, then yeah, onwards!

\cc @sloretz 

> Reviewed 6 of 7 files at r1, 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-:-NP9Uoib0rBV3scE8FLJ:bt7r8q0)* status: 8 of 9 files reviewed, 3 unresolved discussions (waiting on @jwnimmer-tri)

___
*[`bazel_ros2_rules/ros2/ros_cc.bzl` line 132 at r3](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NP9THxkEefHadwCB9Rn:-NP9THxlCVAg_N-Sz7Vk:b-yqnfxs) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/96fa6049a4f6ba453aabfa2637e3dfd796910906/bazel_ros2_rules/ros2/ros_cc.bzl#L132)):*
> ```Starlark
>         name: C/C++ test target name
>         rmw_implementation: optional RMW implementation to run against
>         cc_library_rule: optional cc_library() rule override
> ```

nit Replacing existing line describing `cc_binary_rule` means the signature is now out of sync.

Should either keep that docstring line, or remove it from signature?
___
*[`bazel_ros2_rules/ros2/tools/apply_environment_actions.cc` line 1 at r3](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NP9UghxCqFgdOElaFH3:-NP9UghxCqFgdOElaFH4:bavzrps) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/96fa6049a4f6ba453aabfa2637e3dfd796910906/bazel_ros2_rules/ros2/tools/apply_environment_actions.cc#L1)):*
> ```cpp
> #include "apply_environment_actions.h"
> ```

BTW Is it possible to somehow more directly connect history between this file and `dload_shim`?
___
*[`bazel_ros2_rules/ros2/tools/dload_cc.bzl` line 10 at r3](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NP9TkuJ7AEgBK5mk3V5:-NP9TkuJ7AEgBK5mk3V6:b7tfl1a) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/96fa6049a4f6ba453aabfa2637e3dfd796910906/bazel_ros2_rules/ros2/tools/dload_cc.bzl#L10)):*
> ```Starlark
> 
> _TEMPLATE = """\
> #include "ros2/tools/apply_environment_actions.h"
> ```

Working: From what I remember, we want with the re-`execv` approach because it allows us to honor (silly?) things like `LD_LIBRARY_PATH`.

I fear that this change may defeat that mechanism when it is necessary.
That being said, it could be that Miche and Shane's code (original or updated) already links in all relevant dynamic libraries, thus `LD_LIBRARY_PATH` is not needed.

I believe we should keep `LD_LIBRARY_PATH`, just in case there is a transitive process call that needs it.
(e.g., from a Bazel-linked binary, user calls into a non-rpath binary from their workspace)

If CI passes, then yeah, let's move forward.
If it fails, we may need to re-examine.

Will resolve this once it's clear CI won't fail on this point.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233)* status: 4 of 9 files reviewed, 2 unresolved discussions (waiting on @EricCousineau-TRI)

___
*[`bazel_ros2_rules/ros2/ros_cc.bzl` line 132 at r3](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NP9THxkEefHadwCB9Rn:-NP9cQZ67gScin8WfrqG:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/96fa6049a4f6ba453aabfa2637e3dfd796910906/bazel_ros2_rules/ros2/ros_cc.bzl#L132)):*

nit Replacing existing line describing `cc_binary_rule` means the signature is now out of sync.

Should either keep that docstring line, or remove it from signature?
</blockquote></details>

Done.
___
*[`bazel_ros2_rules/ros2/tools/dload_cc.bzl` line 10 at r3](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NP9TkuJ7AEgBK5mk3V5:-NP9cXTOErQ3jT_82lqi:b-qfxqwi) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/96fa6049a4f6ba453aabfa2637e3dfd796910906/bazel_ros2_rules/ros2/tools/dload_cc.bzl#L10)):*

Working: From what I remember, we want with the re-`execv` approach because it allows us to honor (silly?) things like `LD_LIBRARY_PATH`.

I fear that this change may defeat that mechanism when it is necessary.
That being said, it could be that Miche and Shane's code (original or updated) already links in all relevant dynamic libraries, thus `LD_LIBRARY_PATH` is not needed.

I believe we should keep `LD_LIBRARY_PATH`, just in case there is a transitive process call that needs it.
(e.g., from a Bazel-linked binary, user calls into a non-rpath binary from their workspace)

If CI passes, then yeah, let's move forward.
If it fails, we may need to re-examine.

Will resolve this once it's clear CI won't fail on this point.
</blockquote></details>

Good history, thanks.  I didn't consider LD_LIBRARY_PATH.

I agree with your conclusion, though.  If someone is compiling a program with Bazel, that program itself should never require any LD_LIBRARY_PATH to load correctly.  If it did, then the user has made a mistake in their Bazel rules.  We can  keep the LD_LIBRARY_PATH around to affect subprocesses, though.
___
*[`bazel_ros2_rules/ros2/tools/apply_environment_actions.cc` line 1 at r3](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NP9UghxCqFgdOElaFH3:-NP9btGJ5HsakZ2g9cZG:b-tz03vh) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/96fa6049a4f6ba453aabfa2637e3dfd796910906/bazel_ros2_rules/ros2/tools/apply_environment_actions.cc#L1)):*

BTW Is it possible to somehow more directly connect history between this file and `dload_shim`?
</blockquote></details>

I can just rename it back to the old name.  The "dload" monkier was already vestigial throughout the entire source tree, so it's just as well to keep all of the uses consistently vestigial, to preserve history.

EricCousineau-TRI (approved):
> Reviewed 5 of 7 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233)* status: all files reviewed, 1 unresolved discussion (waiting on @jwnimmer-tri)

___
*[`bazel_ros2_rules/ros2/tools/dload_cc.bzl` line 10 at r3](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NP9TkuJ7AEgBK5mk3V5:-NP9npc1AnqOUx0IN_cQ:bmwnllh) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/96fa6049a4f6ba453aabfa2637e3dfd796910906/bazel_ros2_rules/ros2/tools/dload_cc.bzl#L10)):*
<details><summary><i>Previously, jwnimmer-tri (Jeremy Nimmer) wrote…</i></summary><blockquote>

Good history, thanks.  I didn't consider LD_LIBRARY_PATH.

I agree with your conclusion, though.  If someone is compiling a program with Bazel, that program itself should never require any LD_LIBRARY_PATH to load correctly.  If it did, then the user has made a mistake in their Bazel rules.  We can  keep the LD_LIBRARY_PATH around to affect subprocesses, though.
</blockquote></details>

OK Looks like CI passed for this repo, so will mark this as resolved. Assuming Anzu CI will help us identify other pratfalls.
___
*[`ros2_example_bazel_installed/ros2_example_apps/test/oracle_cc.gdb` line 1 at r5](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NP9o6O49WURw0JviIqG:-NP9o6O49WURw0JviIqH:bvg8ho) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/856ac4e379406e6ed0fc98efea985f45e8c34ce2/ros2_example_bazel_installed/ros2_example_apps/test/oracle_cc.gdb#L1)):*
> ```GDB
> break Oracle::publish_status
> ```

BTW ooh, thanks!
___
*[`ros2_example_bazel_installed/ros2_example_apps/test/oracle_cc.lldb` line 1 at r5](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NP9oA0BCnqrgBes-cyS:-NP9oA0BCnqrgBes-cyT:bgs4tzk) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/bc34b5c59dcafb6040f2896d88536abe7d5f9192/ros2_example_bazel_installed/ros2_example_apps/test/oracle_cc.lldb#L1)):*
> ```
> settings set target.process.stop-on-exec false
> ```

BTW yay!
___
*[`bazel_ros2_rules/ros2/tools/apply_environment_actions.cc` line 1 at r3](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NP9UghxCqFgdOElaFH3:-NP9nyUNDBMi-ZhTDhJS:b-w196av) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/96fa6049a4f6ba453aabfa2637e3dfd796910906/bazel_ros2_rules/ros2/tools/apply_environment_actions.cc#L1)):*
<details><summary><i>Previously, jwnimmer-tri (Jeremy Nimmer) wrote…</i></summary><blockquote>

I can just rename it back to the old name.  The "dload" monkier was already vestigial throughout the entire source tree, so it's just as well to keep all of the uses consistently vestigial, to preserve history.
</blockquote></details>

OK Thanks!

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233)* status: all files reviewed, 2 unresolved discussions (waiting on @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NP9HQ6j2IeLu92z33T7:-NPJobx-8SZavoMXRa5R:bqemqmg) (no related file):*
<details><summary><i>Previously, jwnimmer-tri (Jeremy Nimmer) wrote…</i></summary><blockquote>

Working

Do not merge until Anzu PR 10149 merges first.
</blockquote></details>

Anzu PR 10149 has landed.

An upgrade smoke test in Anzu PR 10155 shows that I still have a little more work to do here first.

___
___
*[`bazel_ros2_rules/ros2/ros_cc.bzl` line 45 at r5](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NPKP3gQ4arJEevGZ6Eb:-NPKP3gQ4arJEevGZ6Ec:b-p0cmlu) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/856ac4e379406e6ed0fc98efea985f45e8c34ce2/bazel_ros2_rules/ros2/ros_cc.bzl#L45)):*
> ```Starlark
>     This wrapper does not support programs that use the 3-argument form of
>     `main` (where environment variables are passed as a third argument).
>     """
> ```

Working

There is one more caveat I'll need to note here.  In case the user's `main` function is defined in a separate library instead of a direct source file, then that library needs to be linked statically.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233)* status: all files reviewed, 2 unresolved discussions (waiting on @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NP9HQ6j2IeLu92z33T7:-NPLpdYF2g12pDjVGR0o:b-lcthkv) (no related file):*
<details><summary><i>Previously, jwnimmer-tri (Jeremy Nimmer) wrote…</i></summary><blockquote>

Anzu PR 10149 has landed.

An upgrade smoke test in Anzu PR 10155 shows that I still have a little more work to do here first.
</blockquote></details>

PR 10155 is passing.  But I still need to merge its prereq 10161 first.

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233)* status: 8 of 9 files reviewed, 1 unresolved discussion (waiting on @EricCousineau-TRI and @jwnimmer-tri)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NP9HQ6j2IeLu92z33T7:-NPNyhrb7S_ar8bl6-SA:b9chmy6) (no related file):*
<details><summary><i>Previously, jwnimmer-tri (Jeremy Nimmer) wrote…</i></summary><blockquote>

PR 10155 is passing.  But I still need to merge its prereq 10161 first.
</blockquote></details>

Everything in Anzu is happy as of last night.  I'm going to do one more rebase there to make sure, and then we should be ready to go here.

___
___
*[`bazel_ros2_rules/ros2/ros_cc.bzl` line 45 at r5](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NPKP3gQ4arJEevGZ6Eb:-NPNygVZ69YWyRIzuJZU:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/856ac4e379406e6ed0fc98efea985f45e8c34ce2/bazel_ros2_rules/ros2/ros_cc.bzl#L45)):*
<details><summary><i>Previously, jwnimmer-tri (Jeremy Nimmer) wrote…</i></summary><blockquote>

Working

There is one more caveat I'll need to note here.  In case the user's `main` function is defined in a separate library instead of a direct source file, then that library needs to be linked statically.
</blockquote></details>

Done.

sloretz (dismissed):
> We can directly replace the user's main(). This simplifies downstream build logic by preserving the invariant that a binary is one program, not two.

If the goal is a single binary, and the wrapper is just setting environment variables, what do you think about making the wrapper a library that has a static global object with a constructor that does the logic? IIUC I think that would allow setting the environment variables before `main()` while only having one binary.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-:-NPO0-X27Rg8fIV_Es1p:b-gu28gt)* status: 8 of 9 files reviewed, 2 unresolved discussions (waiting on @EricCousineau-TRI and @jwnimmer-tri)

___
*[`bazel_ros2_rules/ros2/ros_cc.bzl` line 47 at r6](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NPNzj5aEkW8igokw8mN:-NPNzj5aEkW8igokw8mO:b-lu0mnh) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/da1215d68ee07bdcedcdbd833edcdb120fb3326c/bazel_ros2_rules/ros2/ros_cc.bzl#L47)):*
> ```Starlark
> 
>     This wrapper does not support programs where the main() function is defined
>     in a cc_library. Always pass the main() function definition in `srcs`.
> ```

Does this mean this change won't work with tests that link to `gtest_main`?

https://github.com/RobotLocomotion/drake-ros/blob/94a32508bb7e1e1cfb5c422ac7a64bc836d9dd3d/drake_ros/core/BUILD.bazel#L104-L113

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233)* status: 8 of 9 files reviewed, 3 unresolved discussions (waiting on @EricCousineau-TRI)

___
*[`bazel_ros2_rules/ros2/ros_cc.bzl` line 47 at r6](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NPNzj5aEkW8igokw8mN:-NPOSBy97N_B-tjbN0JL:bsli649) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/da1215d68ee07bdcedcdbd833edcdb120fb3326c/bazel_ros2_rules/ros2/ros_cc.bzl#L47)):*

Does this mean this change won't work with tests that link to `gtest_main`?

https://github.com/RobotLocomotion/drake-ros/blob/94a32508bb7e1e1cfb5c422ac7a64bc836d9dd3d/drake_ros/core/BUILD.bazel#L104-L113

</blockquote></details>

You are probably right that this commit lost the environment variables for all of those googletests.

So now the question is -- why didn't CI detect that things were broken?

sloretz (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233)* status: 8 of 9 files reviewed, 3 unresolved discussions (waiting on @EricCousineau-TRI and @jwnimmer-tri)

___
*[`bazel_ros2_rules/ros2/ros_cc.bzl` line 47 at r6](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NPNzj5aEkW8igokw8mN:-NPOY0fqEEUdY50HbsEW:b-3j6179) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/da1215d68ee07bdcedcdbd833edcdb120fb3326c/bazel_ros2_rules/ros2/ros_cc.bzl#L47)):*
<details><summary><i>Previously, jwnimmer-tri (Jeremy Nimmer) wrote…</i></summary><blockquote>

You are probably right that this commit lost the environment variables for all of those googletests.

So now the question is -- why didn't CI detect that things were broken?
</blockquote></details>

Maybe the only environment variables the tests cared about were LD_LIBRARY_PATH? If so, maybe RPATH is being set such that the tests got access to the libs they needed?

sloretz (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233)* status: 8 of 9 files reviewed, 3 unresolved discussions (waiting on @EricCousineau-TRI and @jwnimmer-tri)

___
*[`bazel_ros2_rules/ros2/ros_cc.bzl` line 47 at r6](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NPNzj5aEkW8igokw8mN:-NPZN3orEQWCbZq3GB3q:bllgpfb) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/da1215d68ee07bdcedcdbd833edcdb120fb3326c/bazel_ros2_rules/ros2/ros_cc.bzl#L47)):*
<details><summary><i>Previously, jwnimmer-tri (Jeremy Nimmer) wrote…</i></summary><blockquote>

I'm thinking along the lines of either

(1) This PR isn't actually broken; the main wrapper still worked for googletests somehow.  (I'll check this for myself locally, later on, to see what's actually happening.)

(2) The environment variables actually never matter anymore, in any situation under any conditions. In that case, we delete all of the shim code and rules entirely.

(3) The environment variable do matter, but our CI test suite didn't notice the regression.  In effect, the shim logic is required for users, but untested in CI.  We need to add more tests in CI to catch this kind of regression.
</blockquote></details>

I think (2) is unlikely. Code accessing the ament index (something querying what plugins are available or what packages exist) will require `AMENT_PREFIX_PATH` to be set.

jwnimmer-tri (commented):
(Still working.)

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-:-NPdXVR69YkEMwfCqB6A:b-65b767)* status: 8 of 10 files reviewed, 2 unresolved discussions (waiting on @EricCousineau-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NPO1NO71tBBTF3aP354:-NPdXBDp7n5_Q2hSdi7w:b-9npk41) (no related file):*
<details><summary><i>Previously, jwnimmer-tri (Jeremy Nimmer) wrote…</i></summary><blockquote>

Working

> ... what do you think about making the wrapper a library that has a static global object with a constructor that does the logic?

That's worth considering.
</blockquote></details>

The challenge here is that we still need `argv[0]` from `main()` to bootstrap the Bazel runfiles index, and we don't have access to that in a global constructor.

I still like the proposal, though, so I'm going to see if I can find a way to thread the needle here.

___
___
*[`bazel_ros2_rules/ros2/ros_cc.bzl` line 47 at r6](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NPNzj5aEkW8igokw8mN:-NPdRJlT5C5w_vsNvQPI:bic9l7d) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/da1215d68ee07bdcedcdbd833edcdb120fb3326c/bazel_ros2_rules/ros2/ros_cc.bzl#L47)):*

I think (2) is unlikely. Code accessing the ament index (something querying what plugins are available or what packages exist) will require `AMENT_PREFIX_PATH` to be set.
</blockquote></details>

It was actually (1) and (3).  The shims _were_ actually still working, but if I purposely broke them (by commenting them out), no tests failed.

Re: (1) -- in particular, the linker `--wrap` **does** work with shared-library googletest main (which is what drake_ros uses), but does **not** work with static-library googletest main (which is what Anzu uses, and needed a tweak to get working correctly).

Re: (3) -- I've added a regression test to this PR now, to prove things are working.

Now with this patch ...

```
--- a/drake_ros/core/BUILD.bazel
+++ b/drake_ros/core/BUILD.bazel
@@ -91,6 +91,7 @@ ros_cc_test(
 ros_cc_test(
     name = "test_drake_ros",
     size = "small",
+    linkstatic = True,
     srcs = ["test/test_drake_ros.cc"],
     rmw_implementation = "rmw_cyclonedds_cpp",
     deps = [
```

... along with the approach in r6, we would catch the failure.

jwnimmer-tri (commented):
@EricCousineau-TRI you should probably cancel your LGTM here, since I'm working on rewriting it.

I've just pushed rough draft to give y'all a flavor of what I'm thinking now.  I'm going to keep testing this locally and in Anzu CI.  I'll let you know when its ready again.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-:-NPx_J5r00h1EWjGbQgW:bifb442)* status: 3 of 13 files reviewed, 2 unresolved discussions (waiting on @EricCousineau-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NPO1NO71tBBTF3aP354:-NPx_WWn07pOk4AawQe0:b-f4eylq) (no related file):*
<details><summary><i>Previously, jwnimmer-tri (Jeremy Nimmer) wrote…</i></summary><blockquote>

The challenge here is that we still need `argv[0]` from `main()` to bootstrap the Bazel runfiles index, and we don't have access to that in a global constructor.

I still like the proposal, though, so I'm going to see if I can find a way to thread the needle here.
</blockquote></details>

New plan: offer a flag for either reexec shim (no change from current) or an opt-in to ldwrap shim (the new way).

Working on it.

___

EricCousineau-TRI (commented):
aye, sgtm :lgtm_cancel: 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-:-NPxbXLIALGE_exraYn6:b-2z89dk)* status: 3 of 13 files reviewed, 2 unresolved discussions (waiting on @jwnimmer-tri)

jwnimmer-tri (commented):
I'm mostly done with code changes now.  (I'm working on to running all of the integration testing.)

I restored the the `dload_shim.cc` code to be closer to what it was previously, so that this PR can focus on the build system changes only.  I suggest comparing to r_base, instead an incremental review.  I plan to re-submit the "C++ platform review" changes there in a follow-up PR.

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-:-NPy2-TeC4dXNC_xwDYT:b-50kcw2)* status: 2 of 15 files reviewed, 1 unresolved discussion (waiting on @EricCousineau-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NPO1NO71tBBTF3aP354:-NPy23pj73DIme-X3LA3:b-896fix) (no related file):*
<details><summary><i>Previously, jwnimmer-tri (Jeremy Nimmer) wrote…</i></summary><blockquote>

New plan: offer a flag for either reexec shim (no change from current) or an opt-in to ldwrap shim (the new way).

Working on it.
</blockquote></details>

Done.

___

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233)* status: 2 of 15 files reviewed, all discussions resolved (waiting on @EricCousineau-TRI)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NP9HQ6j2IeLu92z33T7:-NPy5iWiEQA-HBX92eej:b-p92cgf) (no related file):*
<details><summary><i>Previously, jwnimmer-tri (Jeremy Nimmer) wrote…</i></summary><blockquote>

Everything in Anzu is happy as of last night.  I'm going to do one more rebase there to make sure, and then we should be ready to go here.
</blockquote></details>

Anzu is happy.

___

EricCousineau-TRI (approved):
:lgtm: with some small nits - still fuzzy on chain of events from Anzu PR, but won't block on it.

> Reviewed 6 of 12 files at r8, 7 of 8 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-:-NPyBYiV-d5Ht-w9m8EL:bhm5p16)* status: all files reviewed, 4 unresolved discussions (waiting on @jwnimmer-tri)

___
*[`-- commits` line 4 at r9](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NPy5xf9AefT5UKqluGW):*
nit It's still not 100% clear to me the reason we take this approach in the context of Anzu PR 10046 (upgrading to Bazel 6).

Is this still necessary for the Anzu PR?
If so, is it possible to provide a brief context here?
___
*[`bazel_ros2_rules/ros2/ros_cc.bzl` line 61 at r9](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NPyB9hr8bQgMLhsJDt_:-NPyB9hr8bQgMLhsJDta:bqk7yz2) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/5630252c3c28fc61a16b3a3059d6926ac5bc015e/bazel_ros2_rules/ros2/ros_cc.bzl#L61)):*
> ```Starlark
>       * It does not support programs where the main() function is defined in a
>         cc_library. Always pass the main() function definition in `srcs`.
>         Note that unit test frameworks in particular often a library main().
> ```

nit "often" => "often offer"?
___
*[`bazel_ros2_rules/ros2/ros_cc.bzl` line 190 at r9](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NPy5OVJ5oqI3NxaFIKS:-NPy5OVJ5oqI3NxaFIKT:bopxsbu) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/5630252c3c28fc61a16b3a3059d6926ac5bc015e/bazel_ros2_rules/ros2/ros_cc.bzl#L190)):*
> ```Starlark
>     Additional keyword arguments are forwarded to the `cc_test_rule`.
>     """
>     kwargs.pop("testonly", None)
> ```

nit Motivation for this unclear.

When would someone redundantly use this for a test rule? Should this fail fast if it's set to an unexpected value?
___
*[`bazel_ros2_rules/ros2/tools/dload_shim.cc` line 104 at r9](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NPy4Gcn7JoFLVIx_jsP:-NPy4Gcn7JoFLVIx_jsQ:b-4cj4rr) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/5630252c3c28fc61a16b3a3059d6926ac5bc015e/bazel_ros2_rules/ros2/tools/dload_shim.cc#L104)):*
> ```cpp
>   }
> }
>   
> ```

nit Trailing whitespace.

jwnimmer-tri (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233)* status: 13 of 15 files reviewed, all discussions resolved (waiting on @EricCousineau-TRI)

___
*[`-- commits` line 4 at r9](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NPy5xf9AefT5UKqluGW):*

nit It's still not 100% clear to me the reason we take this approach in the context of Anzu PR 10046 (upgrading to Bazel 6).

Is this still necessary for the Anzu PR?
If so, is it possible to provide a brief context here?
</blockquote></details>

For the genrule in Anzu, we need to execute a ros_cc_binary but we don't have access to its runfiles.  The only way that can work is when it has no runfiles, so therefore it can't have a shim.  That seems like too much detail for the commit message.

EricCousineau-TRI (approved):
> Reviewed 2 of 3 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jwnimmer-tri)

___
*[`-- commits` line 4 at r9](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233#-NPy5xf9AefT5UKqluGW):*
<details><summary><i>Previously, jwnimmer-tri (Jeremy Nimmer) wrote…</i></summary><blockquote>

For the genrule in Anzu, we need to execute a ros_cc_binary but we don't have access to its runfiles.  The only way that can work is when it has no runfiles, so therefore it can't have a shim.  That seems like too much detail for the commit message.
</blockquote></details>

OK SGTM, I think this clarifies it some for me.

EricCousineau-TRI:
Just a small linting issue. I can push fix in next 15min or so if you don't get to it.

jwnimmer-tri:
CI is happy.  PTAL.

EricCousineau-TRI (approved):
> Reviewed 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/233)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jwnimmer-tri)

Issue #229, opened 2023-02-17
cc, py: Should have generic type-casting available for message types


EricCousineau-TRI:
As shown in WIP #169
Relates #215

FYI @adityapande-1995 @sloretz

PR #169, opened 2022-10-21
[WIP] drake_ros_core: Provide better public pybind11 affordances

EricCousineau-TRI:
\cc @sloretz @IanTheEngineer

- - -
This change is [Reviewable](https://reviewable.io/reviews/robotlocomotion/drake-ros/169)

PR #257, opened 2023-03-14
[py] Add generic type casters

adityapande-1995:
This PR adds generic typecasters for geometry messages. For now, I've substituted the individual conversions using a macro, still need to work on fixing the inputs for the macro.

Resolves https://github.com/RobotLocomotion/drake-ros/issues/229

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257)


sloretz (commented):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @adityapande-1995)

___
*[`drake_ros/core/geometry_conversions_pybind.h` line 27 at r1](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NQWgG9t5Xx9HKyCPZSC:-NQWgG9t5Xx9HKyCPZSD:bobj4nc) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/c7b1fcee3d38190250d0d94075303dcbff1d1d1f/drake_ros/core/geometry_conversions_pybind.h#L27)):*
> ```objc
>   \
>       object py_serialize = module::import("rclpy.serialization").attr("serialize_message");\
>       bytes pybytes = py_serialize(source);\
> ```

Before we can serialize, we need to make sure the typesupport has been loaded. That's what the [`check_for_type_support`](https://github.com/RobotLocomotion/drake-ros/blob/a6afd7e0eeca2745d1a6d4319f2a621124a9df8b/drake_ros/drake_ros/core/__init__.py#L53) here does.
___
*[`drake_ros/core/geometry_conversions_pybind.h` line 29 at r1](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NQWdydp3bHSwoEaBLiv:-NQWdydp3bHSwoEaBLiw:bpkz4sp) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/c7b1fcee3d38190250d0d94075303dcbff1d1d1f/drake_ros/core/geometry_conversions_pybind.h#L29)):*
> ```objc
>       bytes pybytes = py_serialize(source);\
>       const std::string content = pybytes;\
>       const auto content_size = content.size() + 1;\
> ```

What's the +1 for? The `\0` byte appended below? If so, it might not be necessary. We don't seem to do it here: https://github.com/RobotLocomotion/drake-ros/blob/a6afd7e0eeca2745d1a6d4319f2a621124a9df8b/drake_ros/drake_ros/core/cc_pybind.cc#L58-L64

adityapande-1995:
*[`drake_ros/core/geometry_conversions_pybind.h` line 29 at r1](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NQWdydp3bHSwoEaBLiv:-NQWsjMSEyc8eEtDWMbX:bo4sncn) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/c7b1fcee3d38190250d0d94075303dcbff1d1d1f/drake_ros/core/geometry_conversions_pybind.h#L29)):*

What's the +1 for? The `\0` byte appended below? If so, it might not be necessary. We don't seem to do it here: https://github.com/RobotLocomotion/drake-ros/blob/a6afd7e0eeca2745d1a6d4319f2a621124a9df8b/drake_ros/drake_ros/core/cc_pybind.cc#L58-L64
</blockquote></details>

Hmm I thought that was more of a safety measure in case the buffers contain garbage data at the end, or is just not null terminated. Given the buffer length specifically, this should not happen, but I see the same used in unit tests here : https://github.com/ros2/rclcpp/blob/33dae5d679751b603205008fcb31755986bcee1c/rclcpp/test/rclcpp/test_serialized_message.cpp#L78-L90. 
On the other hand I could add the same null termination when converting from the C++ type to Python, and that doesn't change things.

adityapande-1995:
*[`drake_ros/core/geometry_conversions_pybind.h` line 27 at r1](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NQWgG9t5Xx9HKyCPZSC:-NQY1k5GDeTAbmi6deYc:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/c7b1fcee3d38190250d0d94075303dcbff1d1d1f/drake_ros/core/geometry_conversions_pybind.h#L27)):*

Before we can serialize, we need to make sure the typesupport has been loaded. That's what the [`check_for_type_support`](https://github.com/RobotLocomotion/drake-ros/blob/a6afd7e0eeca2745d1a6d4319f2a621124a9df8b/drake_ros/drake_ros/core/__init__.py#L53) here does.
</blockquote></details>

Done

adityapande-1995:
*[`drake_ros/core/geometry_conversions_pybind.h` line 29 at r1](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NQWdydp3bHSwoEaBLiv:-NQY1sP61V1gQBO0Btb4:b-hfid7j) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/c7b1fcee3d38190250d0d94075303dcbff1d1d1f/drake_ros/core/geometry_conversions_pybind.h#L29)):*
<details><summary><i>Previously, adityapande-1995 (Aditya Pande) wrote…</i></summary><blockquote>

Hmm I thought that was more of a safety measure in case the buffers contain garbage data at the end, or is just not null terminated. Given the buffer length specifically, this should not happen, but I see the same used in unit tests here : https://github.com/ros2/rclcpp/blob/33dae5d679751b603205008fcb31755986bcee1c/rclcpp/test/rclcpp/test_serialized_message.cpp#L78-L90. 
On the other hand I could add the same null termination when converting from the C++ type to Python, and that doesn't change things.
</blockquote></details>

Added the null termination character for the opposite conversion as well.

EricCousineau-TRI (changes requested):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257)* status: 0 of 1 files reviewed, 6 unresolved discussions (waiting on @adityapande-1995 and @sloretz)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NQzgGeJ0UN7cpW4_4hR:-NQzgwyF5HjNtFZrB-Y0:bbxji8e) (no related file):*

Is there a way to write a type trait (SFINAE) to check if a type is a ROS message IDL type?

If so, I would like us to offer a macro such as `
</blockquote></details>

Ah, incomplete thought:
... macro such as `ROS_MSG_PYBIND_TYPECAST_ALL()`, which will declare a `type_caster<>` that should bind to all type casters.
We should strongly caveat that this should only be used in translation units, not header files.

___

adityapande-1995:

Ah, incomplete thought:
... macro such as `ROS_MSG_PYBIND_TYPECAST_ALL()`, which will declare a `type_caster<>` that should bind to all type casters.
We should strongly caveat that this should only be used in translation units, not header files.
</blockquote></details>

IIUC, do you mean the final usage should be something like :  
```
ROS_MSG_PYBIND_TYPECAST_ALL(geometry_msgs::msg::Quaternion);
ROS_MSG_PYBIND_TYPECAST_ALL(geometry_msgs::msg::Point);
.
.
.
(and so on for all the needed types)
```

The macro would check if it the supplied type is a valid ros msg, extract the text form of the msg type, and invoke the specific ``type_caster<>`` for that type ?

adityapande-1995:
<details><summary><i>Previously, adityapande-1995 (Aditya Pande) wrote…</i></summary><blockquote>

IIUC, do you mean the final usage should be something like :  
```
ROS_MSG_PYBIND_TYPECAST_ALL(geometry_msgs::msg::Quaternion);
ROS_MSG_PYBIND_TYPECAST_ALL(geometry_msgs::msg::Point);
.
.
.
(and so on for all the needed types)
```

The macro would check if it the supplied type is a valid ros msg, extract the text form of the msg type, and invoke the specific ``type_caster<>`` for that type ?

</blockquote></details>

We'll still be supplying the required binding types, right ? I assume theres no way to figure out at compile type what all type castings are required automatically.

EricCousineau-TRI (changes requested):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257)* status: 0 of 1 files reviewed, 6 unresolved discussions (waiting on @adityapande-1995 and @sloretz)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NQzgGeJ0UN7cpW4_4hR:-NR0Pvgf2oy9E19VCo9D:b-osm1vj) (no related file):*
Not exactly.

Users would still (maybe should?) specify `ROS_MSG_PYBIND_TYPECAST(MyMessageType)`
They should then use just `ROS_MSG_PYBIND_TYPECAST()` if they're writing a lot of bindings against message types and want everything to convert via serialization.

> We'll still be supplying the required binding types, right ? 

Not sure if I understand. These macros are meant to supply casting via serilaization, so as long as I have the C++ type and the Python type, I should be able to convert a message completely.
If the C++ type exposes something like `::package_name()` and `::message_name()`, then we can use that to dynamically look up the Python type.

> I assume theres no way to figure out at compile type [...]

Also not sure if I understand.
pybind11 type casting works by effectively calling `type_caster<MyType>` and letting C++ specializations figure out where this ends up.
If there's a specialization that binds to the type (highest affinity), it will use that.
If there's no specialization, it will use `type_caster_generic`, and thus rely on the `py::class_<>` RTTI registry.

`ROS_MSG_PYBIND_TYPECAST(Type)` will provide a specialization for a specific type, and should have highest affinity.
If there is a SFINAE such that we can write `is_ros_message_type_v<Type>`, then we can write `enable_if_t<is_ros_message_type_v<Type>>`, and thus ensure `type_caster<Type>` binds to our serialization caster for all ROS message types.

Does this make sense?

___

EricCousineau-TRI (changes requested):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257)* status: 0 of 1 files reviewed, 7 unresolved discussions (waiting on @adityapande-1995 and @sloretz)

___
*[`drake_ros/core/geometry_conversions_pybind.h` line 39 at r3](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NR0Saoe1X24Y3kD5tkV:-NR0Saoe1X24Y3kD5tkW:b-ny5tcb) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/a788920253f56005b880f75401f70e3c41d5af83/drake_ros/core/geometry_conversions_pybind.h#L39)):*
> ```objc
>       const auto content_size = content.size() + 1;                            \
>                                                                                \
>       rclcpp::SerializedMessage serialized_message(content_size);              \
> ```

When you carve this apart from the macro, can you see if it's easy to factor some of these into functions that can also be used by `PySerializerInterface`?

I had tried to do something to that effect in my prototype #169
Can you take a look there?

adityapande-1995:

Not exactly.

Users would still (maybe should?) specify `ROS_MSG_PYBIND_TYPECAST(MyMessageType)`
They should then use just `ROS_MSG_PYBIND_TYPECAST()` if they're writing a lot of bindings against message types and want everything to convert via serialization.

> We'll still be supplying the required binding types, right ? 

Not sure if I understand. These macros are meant to supply casting via serilaization, so as long as I have the C++ type and the Python type, I should be able to convert a message completely.
If the C++ type exposes something like `::package_name()` and `::message_name()`, then we can use that to dynamically look up the Python type.

> I assume theres no way to figure out at compile type [...]

Also not sure if I understand.
pybind11 type casting works by effectively calling `type_caster<MyType>` and letting C++ specializations figure out where this ends up.
If there's a specialization that binds to the type (highest affinity), it will use that.
If there's no specialization, it will use `type_caster_generic`, and thus rely on the `py::class_<>` RTTI registry.

`ROS_MSG_PYBIND_TYPECAST(Type)` will provide a specialization for a specific type, and should have highest affinity.
If there is a SFINAE such that we can write `is_ros_message_type_v<Type>`, then we can write `enable_if_t<is_ros_message_type_v<Type>>`, and thus ensure `type_caster<Type>` binds to our serialization caster for all ROS message types.

Does this make sense?
</blockquote></details>

Got it, added the ROS_MSG_PYBIND_TYPECAST_ALL() macro. The only caveat is that the second parameter to the PYBIND_TYPE_CASTER() macro, which is supposed to be a docstring for the method, needs to be a constexpr. rosidl does not provide that currently. Ideally the string should be a constexpr form of say "geometry_msgs.msg.Quaternion". I've left it as a blank string for now.

adityapande-1995:
*[`drake_ros/core/geometry_conversions_pybind.h` line 14 at r3](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NQzfQfaE-Dt_i1_ygqR:-NRK7_wKB6O_l42vDnAl:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/a788920253f56005b880f75401f70e3c41d5af83/drake_ros/core/geometry_conversions_pybind.h#L14)):*

Writing longform code in macros is very painful to grok, edit, etc.

We should strive to write this code in actual C++ as much as possible, and have the macro be as small as possible.
Can you write a base type caster to handle this core stuff, add arguments via template arguments, and have the macro just be used for defining the `type_caster<>` specialization?

For example:
https://github.com/RobotLocomotion/drake/blob/v1.14.0/bindings/pydrake/common/wrap_pybind.h#L60-L61
https://github.com/RobotLocomotion/drake/blob/v1.14.0/bindings/pydrake/common/cpp_param_pybind.h#L155-L158
</blockquote></details>

Done.

adityapande-1995:
*[`drake_ros/core/geometry_conversions_pybind.h` line 14 at r3](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NQzgObN99i5rrgH3yno:-NRK7bTB7VmRPE8oJF2g:b-qcnjv1) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/a788920253f56005b880f75401f70e3c41d5af83/drake_ros/core/geometry_conversions_pybind.h#L14)):*

nit This macro does not indicate usage (for pybind11). This should be qualified, e.g.

```
ROS_MSG_PYBIND_TYPECAST(...)
```
</blockquote></details>

Changed to the TYPECAST_ALL style macro.

adityapande-1995:
*[`drake_ros/core/geometry_conversions_pybind.h` line 14 at r3](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NQzgXD1CNmsRcMgLIgH:-NRK7jxc8SNvP3vc7jzW:bmeisxw) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/a788920253f56005b880f75401f70e3c41d5af83/drake_ros/core/geometry_conversions_pybind.h#L14)):*

nit Does ROS IDL offer static / constexpr access to package / subdir / message name as strings?

If so, ideally we extract that from type traits, rather than requiring users to decompose it in the macro itself.
e.g.
https://github.com/RobotLocomotion/drake/blob/v1.14.0/bindings/pydrake/systems/lcm_pybind.h#L32-L33
</blockquote></details>

I used the rosidl type traits for the actual conversions, but couldn't use for the docstring. Left that section as a black for now.

adityapande-1995:
> > _[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257)_ status: 0 of 1 files reviewed, 6 unresolved discussions (waiting on @adityapande-1995 and @sloretz)
> 
> _[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NQzgGeJ0UN7cpW4_4hR:-NQzgwyF5HjNtFZrB-Y0:bbxji8e) (no related file):_
> 
> Previously, EricCousineau-TRI (Eric Cousineau) wrote…
> Ah, incomplete thought: ... macro such as `ROS_MSG_PYBIND_TYPECAST_ALL()`, which will declare a `type_caster<>` that should bind to all type casters. We should strongly caveat that this should only be used in translation units, not header files.

Do we need a macro in this case ? We could just have the code section as it is.

adityapande-1995:
Converting to a draft until a couple of things are addressed : 
- Splitting serializing and deserializing into smaller functions
- Do we need a macro at all ? 
- Is it okay to ignore the docstring in ``PYBIND11_TYPE_CASTER()`` macro ?

adityapande-1995:
*[`drake_ros/core/geometry_conversions_pybind.h` line 18 at r4](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NRYCTAY6IS6RSyXY5nk:-NR_aVFPD5a5On2poeoL:b-kcwu2i) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/d5f574bc41fd082cad8abd871db933ca13718c37/drake_ros/core/geometry_conversions_pybind.h#L18)):*

Sorry for not being clearer, but I think we should provide users two entry points:
- The `*_ALL()` flavor
- A message-specific flavor

The guts of the ROS 2 IDL <-> pybind11 should happen inside of our own namespace `drake_ros::core`, e.g.

Does this make sense?
</blockquote></details>

Split the namespaces, done.

adityapande-1995:
*[`drake_ros/core/geometry_conversions_pybind.h` line 38 at r4](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NRYC2gtB0xYWI7j0xz-:-NR_a_yaAOENfPjriRjO:bzcbmt1) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/d5f574bc41fd082cad8abd871db933ca13718c37/drake_ros/core/geometry_conversions_pybind.h#L38)):*

For succinctness / clarity of authoring, most of our code should live in our own namespace.
We should then make thin shells in `pybind11::detail` that shell out to our code.
</blockquote></details>

Done, split namespaces.

adityapande-1995:
Hi @EricCousineau-TRI  ! 

Yes, splitting the functions into separate ones is done. One question : Now that we have both macros, how do you want the test to work ? i.e. The typecaster ultimately is a part of cc_pybind.cc (and exposed using ``import drake_ros.core``). IIUC, we want a the python test to run twice : once using the ``ALL()`` typecaster, and once using the per msg typecasters. So ultimately, these 2 sections are equivalent as far as the test is concerned : 

In ``ros_idl_pybind.h`` : 
```
// Generic typecaster for all ROS 2 messages.
ROS_MSG_PYBIND_TYPECAST_ALL();
```

In ``geometry_conversions_pybind.h`` :
```
// Generic typecaster for specific ROS 2 messages.
// This method can be used instead of the ROS_MSG_PYBIND_TYPECAST_ALL() macro.
// ROS_MSG_PYBIND_TYPECAST(geometry_msgs::msg::Quaternion);
// ROS_MSG_PYBIND_TYPECAST(geometry_msgs::msg::Point);
// ROS_MSG_PYBIND_TYPECAST(geometry_msgs::msg::Vector3);
// ROS_MSG_PYBIND_TYPECAST(geometry_msgs::msg::Twist);
// ROS_MSG_PYBIND_TYPECAST(geometry_msgs::msg::Accel);
// ROS_MSG_PYBIND_TYPECAST(geometry_msgs::msg::Wrench);
// ROS_MSG_PYBIND_TYPECAST(geometry_msgs::msg::Pose);
// ROS_MSG_PYBIND_TYPECAST(geometry_msgs::msg::Transform);

```

Do you want like an ``#ifdef`` flag in ``cc_pybind.cc`` that selects the appropriate ``#include`` during compile time ? So by default it can go with the ``ALL()`` style, and if we set the flag it goes for the per msg style.

adityapande-1995 (commented):
For the constexpr name, this is the generated function for say the ``Quaternion`` msg: 
```
template<>
inline const char * name<geometry_msgs::msg::Quaternion>()
{
  return "geometry_msgs/msg/Quaternion";
}
```

This is in the generated file ``quaternion__traits.hpp``. You can do a ``find -L . | grep quaternon__traits`` inside ``drake_ros`` after it is built. The compiler complains that ``const char*`` cannot be converted to a ``const char[]``. 

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-:-NR_gpf93UHM1_sclw5q:bsf20w)* status: 0 of 3 files reviewed, 8 unresolved discussions (waiting on @EricCousineau-TRI and @sloretz)

EricCousineau-TRI (commented):
> Reviewed 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257)* status: 1 of 3 files reviewed, 6 unresolved discussions (waiting on @adityapande-1995 and @sloretz)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NQzgGeJ0UN7cpW4_4hR:-NRkAFi5-ESKG3YtNldJ:b8sq2x0) (no related file):*

> For the constexpr name, this is the generated function for say the Quaternion msg:

Yes, that makes sense that `const char*` cannot be used in a truly `contexpr` sense.
It seems defective to not provide contexpr access to generated types, though.

Can you file an upstream issue, and track it here via TODO?

___
*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NRYDy_19F-OT_tLeyGj:-NRkAKpj9V707BwPBX13:b-2jzec1) (no related file):*
> One question : Now that we have both macros, how do you want the test to work ?

Let's keep it simple. `ros_idl_pybind.h` only declares the macros, it does not use them.
`geometry_conversions_pybind.h` only uses the per-type conversion marcos.

For testing, we should just create two entirely separate translation units `(*_pybind.cc` or `_py.cc`).
One that briefly tests a per-message macro, one that tests the `*_ALL()` macro.

___
___
*[`drake_ros/core/geometry_conversions_pybind.h` line 1 at r4](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NRYDpKT19B1_UtRiscS:-NRk9Tag8YhH-IjbqntC:buz0qgb) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/d5f574bc41fd082cad8abd871db933ca13718c37/drake_ros/core/geometry_conversions_pybind.h#L1)):*
> You should still make a header, geometry_conversions_pybind.h, that would do per-message declaration.

Can I ask if you can address this?
___
*[`drake_ros/core/ros_idl_pybind.h` line 21 at r6](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NRk977xCgIAh5JKiPEv:-NRk977y6pytEv1Iky1r:b7gb49s) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/750cf81c37378a65529cab0f309c8320f259fcc8/drake_ros/core/ros_idl_pybind.h#L21)):*
> ```objc
>   template <typename T>                                                     \
>   struct type_caster<                                                       \
>       T, std::enable_if_t<rosidl_generator_traits::is_message<T>::value>> { \
> ```

You should be able to inherit form another typecaster; that should reduce the repeated code.
See prior example pointed out:
https://github.com/RobotLocomotion/drake/blob/v1.14.0/bindings/pydrake/common/cpp_param_pybind.h#L155-L158

adityapande-1995:


> For the constexpr name, this is the generated function for say the Quaternion msg:

Yes, that makes sense that `const char*` cannot be used in a truly `contexpr` sense.
It seems defective to not provide contexpr access to generated types, though.

Can you file an upstream issue, and track it here via TODO?
</blockquote></details>

Done

adityapande-1995:
*[`drake_ros/core/ros_idl_pybind.h` line 21 at r6](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NRk977xCgIAh5JKiPEv:-NRtm6iS5-FB4yHc7kyu:b-896fjb) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/750cf81c37378a65529cab0f309c8320f259fcc8/drake_ros/core/ros_idl_pybind.h#L21)):*

You should be able to inherit form another typecaster; that should reduce the repeated code.
See prior example pointed out:
https://github.com/RobotLocomotion/drake/blob/v1.14.0/bindings/pydrake/common/cpp_param_pybind.h#L155-L158
</blockquote></details>

Done

adityapande-1995:
*[`drake_ros/core/geometry_conversions_pybind.h` line 1 at r4](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NRYDpKT19B1_UtRiscS:-NRtoGr5DowRzTcyOdsL:bmon52i) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/d5f574bc41fd082cad8abd871db933ca13718c37/drake_ros/core/geometry_conversions_pybind.h#L1)):*

> You should still make a header, geometry_conversions_pybind.h, that would do per-message declaration.

Can I ask if you can address this?
</blockquote></details>

Added a ``geometry_conversions_pybind.h`` that uses the message specific declaration.

adityapande-1995:

> One question : Now that we have both macros, how do you want the test to work ?

Let's keep it simple. `ros_idl_pybind.h` only declares the macros, it does not use them.
`geometry_conversions_pybind.h` only uses the per-type conversion marcos.

For testing, we should just create two entirely separate translation units `(*_pybind.cc` or `_py.cc`).
One that briefly tests a per-message macro, one that tests the `*_ALL()` macro.
</blockquote></details>

Done !

EricCousineau-TRI (changes requested):
> Reviewed 8 of 8 files at r7, all commit messages.
> Dismissed @adityapande-1995 and @sloretz from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257)* status: all files reviewed, 16 unresolved discussions (waiting on @adityapande-1995)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NRYDy_19F-OT_tLeyGj:-NSrlZWu6Kn1J3n2ygn_:bo9v0c4) (no related file):*
<details><summary><i>Previously, adityapande-1995 (Aditya Pande) wrote…</i></summary><blockquote>

Done !
</blockquote></details>

Looks better! some remaining comments, PTAL!

___
___
*[`drake_ros/core/geometry_conversions_pybind.h` line 1 at r4](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrl7bQ-jQqfg-HuNFd:-NSrl7bQ-jQqfg-HuNFe:b3e073d) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/d5f574bc41fd082cad8abd871db933ca13718c37/drake_ros/core/geometry_conversions_pybind.h#L1)):*
> ```objc
> #include <regex>
> ```

nit Apologies for back and forth, but this filename should be more consistent with the macros and types.

Options:
- Rename this to `ros_message_pybind.h` to align w/ `ros_message_type_caster`
- Rename macros and types to something like `ros_idl_type_caster`, and rename stuff to use `ros_idl_type_caster` in the name.

IDL may be a bit hard for users to discover, so maybe better to use `ros_message`?
___
*[`drake_ros/core/geometry_conversions_pybind.h` line 8 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrk2mN9gE3YJ0YU2eq:-NSrk2mN9gE3YJ0YU2er:b-eq5ktv) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/core/geometry_conversions_pybind.h#L8)):*
> ```objc
> // Generic typecaster for specific ROS 2 messages.
> // This method can be used instead of the ROS_MSG_PYBIND_TYPECAST_ALL() macro.
> ROS_MSG_PYBIND_TYPECAST(geometry_msgs::msg::Quaternion);
> ```

These macros should be called at top-level scope.

For example, it should function as `PYBIND11_MAKE_OPAQUE` does.
https://github.com/pybind/pybind11/blob/ed466da571fbc1d711351eb818e4bf82adb99eca/include/pybind11/cast.h#L1659-L1665
Or `PYBIND11_NUMPY_DTYPE`
https://github.com/pybind/pybind11/blob/ed466da571fbc1d711351eb818e4bf82adb99eca/include/pybind11/numpy.h#L1509-L1512
___
*[`drake_ros/drake_ros/BUILD.bazel` line 35 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrg-m8Dp7jCUJpu3jF:-NSrg-m90SzjrIRw20Hl:bg4poki) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/drake_ros/BUILD.bazel#L35)):*
> ```Starlark
> )
> 
> pybind_py_library(
> ```

nit Purpose of this binding library unclear.

Can you place this closer to `generic_typecaster_test`, and perhaps change name to something as follows?
`ros_message_type_caster_test_cc_py`

You should also adjust `cc_so_name = "ros_message_type_caster_test_cc"`
___
*[`drake_ros/drake_ros/BUILD.bazel` line 41 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrfsFV2b8pGyK42rVa:-NSrfsFV2b8pGyK42rVb:b-8qxaui) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/drake_ros/BUILD.bazel#L41)):*
> ```Starlark
>         "//:drake_ros_shared_library",
>     ],
>     cc_so_name = "_cc_generic_typecaster",
> ```

nit This should be `testonly = 1`.

Can you add that attribute here?
___
*[`drake_ros/drake_ros/BUILD.bazel` line 43 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrhGxB9xiC7uJrvxmr:-NSrhGxC4cESXeEk_0at:brk794n) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/drake_ros/BUILD.bazel#L43)):*
> ```Starlark
>     cc_so_name = "_cc_generic_typecaster",
>     cc_srcs = [
>         "cc_py_generic_typecaster.cc",
> ```

nit These functions are for testing.

Can you place them in the appropriate `test/` folder, and collocate them?
___
*[`drake_ros/drake_ros/BUILD.bazel` line 128 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrkzYc9jX8Npr6fxuS:-NSrkzYd5xapxk5iAHu1:b-wcyrb) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/drake_ros/BUILD.bazel#L128)):*
> ```Starlark
> 
> ros_py_test(
>     name = "generic_typecaster_test",
> ```

nit Name `generic_typecaster` is unclear.

I recommend renaming the test to something more specific, e.g. `ros_message_type_caster_test`
___
*[`drake_ros/drake_ros/cc_py_generic_typecaster.cc` line 14 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrjWgw0cby4oM0Pr2t:-NSrjWgx3P0JpbSlujcb:bn7kzkt) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/drake_ros/cc_py_generic_typecaster.cc#L14)):*
> ```cpp
>   m.doc() = "Mock python bindings for drake_ros";
> 
>   DefTestTypecasting(m);
> ```

This should have two translation units, and two functions. It should have functions that looks something like:
```
DefTypecastingViaSpecificTypes(m);
DefTypecastingViaAll(m);
```

I would recommend renaming those translation units to
```
test/ros_message_type_caster_test_via_specific_types_pybind.cc
test/ros_message_type_caster_test_via_all_pybind.cc
```
___
*[`drake_ros/drake_ros/core/cc_pybind_mock.cc` line 20 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrgFLA5KzPPSI2wPgg:-NSrgFLA5KzPPSI2wPgh:bt9p5un) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/drake_ros/core/cc_pybind_mock.cc#L20)):*
> ```cpp
> namespace mock_test {
> template <typename T>
> T testTypecasting(T msg) {
> ```

nit Per GSG, can you use `CamelCase` here?
___
*[`drake_ros/drake_ros/core/cc_pybind_mock.cc` line 38 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrgI-wD-MyAevkS-gf:-NSrgI-x9tYdgf6kEe8s:bj3c12p) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/drake_ros/core/cc_pybind_mock.cc#L38)):*
> ```cpp
>   py::class_<DrakeRos>(m, "DrakeRos");
> 
>   m.def("testTypecasting",
> ```

nit Same as above.
___
*[`drake_ros/drake_ros/test/generic_typecaster_test.py` line 10 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrgzCC9qyaMoy5XF9L:-NSrgzCC9qyaMoy5XF9M:brewh1j) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/drake_ros/test/generic_typecaster_test.py#L10)):*
> ```Python
> from geometry_msgs.msg import Polygon, Point32, Quaternion
> 
> # This message type is not covered by geometry_messages_pybind.h
> ```

nit Can you move these comments to the function, using docstring `""" ... """`?
___
*[`drake_ros/drake_ros/test/generic_typecaster_test.py` line 27 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrnlKyEIpUkNakTUb5:-NSrnlKyEIpUkNakTUb6:b-jv19u2) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/drake_ros/test/generic_typecaster_test.py#L27)):*
> ```Python
> # but since the module is not included, this should still be
> # processed using the generic typecaster.
> def test_msg_quaternion():
> ```

Can you explicitly test what an incorrect / mismsatched type looks like?

Ideally, the error message should provide the user with enough information so they can kind of figure out what's going on.
___
*[`drake_ros/core/ros_idl_pybind.h` line 26 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrmSfI9t-qhyzZoOac:-NSrmSfI9t-qhyzZoOad:bkzdyl0) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/core/ros_idl_pybind.h#L26)):*
> ```objc
> 
> // Generic (C++ <-> Python) typecaster for a specific ROS 2 message.
> #define ROS_MSG_PYBIND_TYPECAST(T) \
> ```

Single-message declaration should ideally declare a name.

Can you change the macro to look something like this?
```
ROS_MESSAGE_PYBIND_TYPECASTER(T) \
   namespace pybind11 { namespace detail {
   template <> \
   struct type_caster<T> : public ros_message_type_caster<T> { \
     public: \
       static constexpr auto name = _("TypeCaster[" #T "]"); \
   } \
```

When pybind11 error messages crop up, then they should see something like:
```
incompatible function arguments... The following types are supported:
   1. (msg: RosTypeCaster[geometry_msgs::Vector3])
```
___
*[`drake_ros/core/ros_idl_pybind.h` line 113 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrmDgU5xiEDUZKtihC:-NSrmDgVAIk9cnhilYnN:b-3w90us) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/core/ros_idl_pybind.h#L113)):*
> ```objc
> namespace detail {
> 
> // TODO(Aditya): Apply constexpr message name to type caster macro
> ```

nit Can you move this TODO closer to the usage of the name for `PYBIND11_TYPE_CASTER`?
___
*[`drake_ros/core/ros_idl_pybind.h` line 116 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrhZIlESHU-Vr8Hsx1:-NSrhZIlESHU-Vr8Hsx2:bz1zwcz) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/core/ros_idl_pybind.h#L116)):*
> ```objc
> // after https://github.com/ros2/rosidl/issues/734 is resolved.
> template <typename T>
> struct generic_type_caster {
> ```

Name unclear. Should be qualified by `ros_message` somewhere in name.

Consider `ros_message_type_caster`.
___
*[`drake_ros/core/ros_idl_pybind.h` line 116 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrjEJh02QMh3bTTjhP:-NSrjEJh02QMh3bTTjhQ:b-8dxvwz) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/core/ros_idl_pybind.h#L116)):*
> ```objc
> // after https://github.com/ros2/rosidl/issues/734 is resolved.
> template <typename T>
> struct generic_type_caster {
> ```

This caster type should live in `drake_ros::drake_ros_py`.

Can you move it there?
___
*[`drake_ros/core/ros_idl_pybind.h` line 118 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrlrxAEIBdDS3nFtvk:-NSrlrxAEIBdDS3nFtvl:bu168oo) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/core/ros_idl_pybind.h#L118)):*
> ```objc
> struct generic_type_caster {
>  public:
>   PYBIND11_TYPE_CASTER(T, py::detail::_(""));
> ```

nit Blank string for type will make it very hard to reason about error messages.

Can you perhaps put something like `RosTypeCaster[Unknown]` for the name, so users can search for it?

adityapande-1995:
*[`drake_ros/drake_ros/core/cc_pybind_mock.cc` line 38 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrgI-wD-MyAevkS-gf:-NTGbQXmCIIn3EDQnS-a:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/drake_ros/core/cc_pybind_mock.cc#L38)):*

nit Same as above.
</blockquote></details>

Done

adityapande-1995:
*[`drake_ros/drake_ros/core/cc_pybind_mock.cc` line 20 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrgFLA5KzPPSI2wPgg:-NTGbRk9DI1KYnY3jeK-:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/drake_ros/core/cc_pybind_mock.cc#L20)):*

nit Per GSG, can you use `CamelCase` here?
</blockquote></details>

Done

adityapande-1995:
*[`drake_ros/drake_ros/test/generic_typecaster_test.py` line 10 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrgzCC9qyaMoy5XF9L:-NTGbU7M0CT07AThF36X:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/drake_ros/test/generic_typecaster_test.py#L10)):*

nit Can you move these comments to the function, using docstring `""" ... """`?
</blockquote></details>

Done

adityapande-1995:
*[`drake_ros/drake_ros/BUILD.bazel` line 128 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrkzYc9jX8Npr6fxuS:-NTGbX4N1OG_IMIPmBSG:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/drake_ros/BUILD.bazel#L128)):*

nit Name `generic_typecaster` is unclear.

I recommend renaming the test to something more specific, e.g. `ros_message_type_caster_test`
</blockquote></details>

Done

adityapande-1995:
*[`drake_ros/drake_ros/BUILD.bazel` line 43 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrhGxB9xiC7uJrvxmr:-NTGbZDJEiF8CW_a_zIn:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/drake_ros/BUILD.bazel#L43)):*

nit These functions are for testing.

Can you place them in the appropriate `test/` folder, and collocate them?
</blockquote></details>

Done

adityapande-1995:
*[`drake_ros/drake_ros/BUILD.bazel` line 41 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrfsFV2b8pGyK42rVa:-NTGb_dhCj2cB5C_B0ML:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/drake_ros/BUILD.bazel#L41)):*

nit This should be `testonly = 1`.

Can you add that attribute here?
</blockquote></details>

Done

adityapande-1995:
*[`drake_ros/drake_ros/BUILD.bazel` line 35 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrg-m8Dp7jCUJpu3jF:-NTGbb7Q8AZLU1S9v-c6:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/drake_ros/BUILD.bazel#L35)):*

nit Purpose of this binding library unclear.

Can you place this closer to `generic_typecaster_test`, and perhaps change name to something as follows?
`ros_message_type_caster_test_cc_py`

You should also adjust `cc_so_name = "ros_message_type_caster_test_cc"`
</blockquote></details>

Done

adityapande-1995:
*[`drake_ros/core/ros_idl_pybind.h` line 118 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrlrxAEIBdDS3nFtvk:-NTGbd9mCb6mIZok4AgB:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/core/ros_idl_pybind.h#L118)):*

nit Blank string for type will make it very hard to reason about error messages.

Can you perhaps put something like `RosTypeCaster[Unknown]` for the name, so users can search for it?
</blockquote></details>

Done.

adityapande-1995:
*[`drake_ros/core/ros_idl_pybind.h` line 116 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrjEJh02QMh3bTTjhP:-NTGbf0u-VINrfMYM6vA:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/core/ros_idl_pybind.h#L116)):*

This caster type should live in `drake_ros::drake_ros_py`.

Can you move it there?
</blockquote></details>

Done

adityapande-1995:
*[`drake_ros/core/ros_idl_pybind.h` line 113 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrmDgU5xiEDUZKtihC:-NTGbll7-0hwHMIiTgjQ:bb6un1j) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/core/ros_idl_pybind.h#L113)):*

nit Can you move this TODO closer to the usage of the name for `PYBIND11_TYPE_CASTER`?
</blockquote></details>

Done

adityapande-1995:
*[`drake_ros/core/ros_idl_pybind.h` line 26 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrmSfI9t-qhyzZoOac:-NTGbpXK3HVaL0oplsIf:b-kobypd) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/core/ros_idl_pybind.h#L26)):*

Single-message declaration should ideally declare a name.

Can you change the macro to look something like this?
```
ROS_MESSAGE_PYBIND_TYPECASTER(T) \
   namespace pybind11 { namespace detail {
   template <> \
   struct type_caster<T> : public ros_message_type_caster<T> { \
     public: \
       static constexpr auto name = _("TypeCaster[" #T "]"); \
   } \
```

When pybind11 error messages crop up, then they should see something like:
```
incompatible function arguments... The following types are supported:
   1. (msg: RosTypeCaster[geometry_msgs::Vector3])
```
</blockquote></details>

How should we handle this now the we are deriving from a common struct for the typecaster ?

adityapande-1995:
*[`drake_ros/drake_ros/cc_py_generic_typecaster.cc` line 14 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrjWgw0cby4oM0Pr2t:-NTGbzwm5Bip0YFj_0lF:b-9e000k) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/drake_ros/cc_py_generic_typecaster.cc#L14)):*

This should have two translation units, and two functions. It should have functions that looks something like:
```
DefTypecastingViaSpecificTypes(m);
DefTypecastingViaAll(m);
```

I would recommend renaming those translation units to
```
test/ros_message_type_caster_test_via_specific_types_pybind.cc
test/ros_message_type_caster_test_via_all_pybind.cc
```
</blockquote></details>

So in the module ultimately, we'll have both the specific and the generic typecaster at the same time ? How do we know if the module is internally using the all or the specific typecasters in that case ? The actual drake_ros geometry conversion has test for the specific types. I mean, in the test here :

```
import drake_ros.ros_message_all_typecaster_test_module as all_typecaster

def foo():
  assert all_typecaster.TestTypecasting(.....) == expected_msg
```

this assumes the module only contains the all style typecaster. Should we separate the modules if we want to make that clear ?

jwnimmer-tri:
BTW @adityapande-1995 when using Reviewable, there is a big green "Publish" in the upper right, which publishes _all_ of your responses in a batch.  The little image in the corner of each discussion thread should generally not be used, as it sends an email all on its own, leading to an email flood.

adityapande-1995:
*[`drake_ros/drake_ros/cc_py_generic_typecaster.cc` line 14 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrjWgw0cby4oM0Pr2t:-NTLDFQt7x3aRiZjWC95:b762b5u) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/drake_ros/cc_py_generic_typecaster.cc#L14)):*
<details><summary><i>Previously, adityapande-1995 (Aditya Pande) wrote…</i></summary><blockquote>

So in the module ultimately, we'll have both the specific and the generic typecaster at the same time ? How do we know if the module is internally using the all or the specific typecasters in that case ? The actual drake_ros geometry conversion has test for the specific types. I mean, in the test here :

```
import drake_ros.ros_message_all_typecaster_test_module as all_typecaster

def foo():
  assert all_typecaster.TestTypecasting(.....) == expected_msg
```

this assumes the module only contains the all style typecaster. Should we separate the modules if we want to make that clear ? 
</blockquote></details>

I added a temp commit here : https://github.com/RobotLocomotion/drake-ros/pull/257/commits/918db4b3c0567896b1d4f651530a5b6f631c0a8a which adds 2 separate translation units, one with the all-style typecaster and one with the specific ones. For the test, the Polygon msg typecaster is *not* declared using the specific types, and the corresponding test should fail. Instead, see the test pass, probably using the all-typecaster internally. That is, in this arrangement, we can't have a test that would fail for specific typecaster but pass for the all() one.

Just added this temp commit to demonstrate, I can revert if we can't resolve this.

Internally, we're probably polluting the py::detail namespace with both typecasters, so AFA the python module is concerned, the translation units are not *separating* the typecasters.

EricCousineau-TRI (changes requested):
> Reviewed 1 of 5 files at r8, 5 of 5 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257)* status: all files reviewed, 4 unresolved discussions (waiting on @adityapande-1995 and @sloretz)

___
*[`drake_ros/core/geometry_conversions_pybind.h` line 8 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrk2mN9gE3YJ0YU2eq:-NYUUeS90iSO6RZXn--Q:b-bqyqfp) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/core/geometry_conversions_pybind.h#L8)):*

Are you saying the `namespace ...` declarations should live inside the macros?
</blockquote></details>

Sorry for the delay - yup!
___
*[`drake_ros/drake_ros/test/ros_message_type_caster_test_via_specific_types_pybind.cc` line 36 at r9](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NYUVOCM6wsn9sJlQjuU:-NYUVOCM6wsn9sJlQjuV:brkq2n1) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/918db4b3c0567896b1d4f651530a5b6f631c0a8a/drake_ros/drake_ros/test/ros_message_type_caster_test_via_specific_types_pybind.cc#L36)):*
> ```cpp
>   m.doc() = "Python bindings for testing the generic typecaster";
> 
>   // py::class_<DrakeRos>(m, "DrakeRos");
> ```

nit Unused?
___
*[`drake_ros/drake_ros/cc_py_generic_typecaster.cc` line 14 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrjWgw0cby4oM0Pr2t:-NYUVXUA1jLjgVQfUtGL:b-mio3ft) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/drake_ros/cc_py_generic_typecaster.cc#L14)):*

Based on your findings Aditya, it sounds we can't be sure the test is actually using the typecaster from the "type specific" macro when the "alll types" macro is used, even if it was used in a different translation unit. I'd recommend moving forward with tests for each macro being in different CPython extensions.
</blockquote></details>

Working: Hm... that doesn't sound right. `type_caster<T>` for a non-registered type should only be able to bind through a translation unit.
Will briefly test that out.
___
*[`drake_ros/drake_ros/test/generic_typecaster_test.py` line 27 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrnlKyEIpUkNakTUb5:-NYUVAKG20Y5H2j6o4A5:b77yfue) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/drake_ros/test/generic_typecaster_test.py#L27)):*

I think we get that for free from pytest. IIRC it will print the values of all parts of the expression used in the `assert` statement.
</blockquote></details>

OK SGTM for now!

EricCousineau-TRI (changes requested):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257)* status: all files reviewed, 5 unresolved discussions (waiting on @adityapande-1995 and @sloretz)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NYU_48Q-6nw0Z0x0ds-:-NYU_48Q-6nw0Z0x0ds0:bdxlsth) (no related file):*
Working: I have a few other fixes incoming.

___

EricCousineau-TRI (changes requested):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257)* status: all files reviewed, 8 unresolved discussions (waiting on @adityapande-1995 and @sloretz)

___
*[`drake_ros/core/ros_idl_pybind.h` line 30 at r9](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NYU_ywbC44b5-Rd3x03:-NYU_ywc4RsOnWFo5VPI:b-ugtfzw) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/918db4b3c0567896b1d4f651530a5b6f631c0a8a/drake_ros/core/ros_idl_pybind.h#L30)):*
> ```objc
>   struct type_caster<T> : public drake_ros::drake_ros_py::ros_message_type_caster<T> {};
> 
> namespace drake_ros {
> ```

nit This file should ideally only define things in `drake_ros::drake_ros_py`.
___
*[`drake_ros/core/ros_idl_pybind.h` line 112 at r9](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NYU_nyh47pVLw2aa3P_:-NYU_nyh47pVLw2aa3Pa:b-x15tmf) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/918db4b3c0567896b1d4f651530a5b6f631c0a8a/drake_ros/core/ros_idl_pybind.h#L112)):*
> ```objc
> namespace drake_ros {
> namespace drake_ros_py {
> using namespace std;
> ```

We should not do `using namespace` if it can be avoided (and I believe we can avoid it).
___
*[`drake_ros/core/ros_idl_pybind.h` line 113 at r9](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NYU_snt2PfkiquocnWu:-NYU_snu3gGd6wkq-LTw:b1p9azv) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/918db4b3c0567896b1d4f651530a5b6f631c0a8a/drake_ros/core/ros_idl_pybind.h#L113)):*
> ```objc
> namespace drake_ros_py {
> using namespace std;
> using handle = py::handle;
> ```

nit These other using statemetns can be placed in a macro.

EricCousineau-TRI (changes requested):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257)* status: all files reviewed, 8 unresolved discussions (waiting on @adityapande-1995 and @sloretz)

___
*[`drake_ros/drake_ros/cc_py_generic_typecaster.cc` line 14 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrjWgw0cby4oM0Pr2t:-NYUczaqDWzy7OpVzzOj:b-a6mfjh) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/drake_ros/cc_py_generic_typecaster.cc#L14)):*

Working: Hm... that doesn't sound right. `type_caster<T>` for a non-registered type should only be able to bind through a translation unit.
Will briefly test that out.
</blockquote></details>

Yeah, I'm finding the same behavior. In this case, we should just create separate modules.

EricCousineau-TRI (changes requested):
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257)* status: 3 of 12 files reviewed, 4 unresolved discussions (waiting on @adityapande-1995 and @sloretz)

___
*[`.gitignore` line 2 at r10](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NYUgMub79c8yNqvRFyf:-NYUgMub79c8yNqvRFyg:b-gsjm6i) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/d1c7fad2d29c6c0962235d0f17b49fc5a329427a/.gitignore#L2)):*
> ```ignore
> bazel-*
> /.vscode/
> ```

Working: Oops should remove this.
___
*[`drake_ros/core/geometry_conversions_pybind.h` line 8 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrk2mN9gE3YJ0YU2eq:-NYUgfltE1G5z8YkaZEw:blwrh79) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/core/geometry_conversions_pybind.h#L8)):*

Sorry for the delay - yup!
</blockquote></details>

Working: Will fix.
___
*[`drake_ros/core/ros_idl_pybind.h` line 30 at r9](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NYU_ywbC44b5-Rd3x03:-NYUgPVd6_fJLC--wmKA:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/918db4b3c0567896b1d4f651530a5b6f631c0a8a/drake_ros/core/ros_idl_pybind.h#L30)):*

nit This file should ideally only define things in `drake_ros::drake_ros_py`.
</blockquote></details>

Done.
___
*[`drake_ros/core/ros_idl_pybind.h` line 112 at r9](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NYU_nyh47pVLw2aa3P_:-NYUgQGB8Kjj6p-IMaSB:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/918db4b3c0567896b1d4f651530a5b6f631c0a8a/drake_ros/core/ros_idl_pybind.h#L112)):*

We should not do `using namespace` if it can be avoided (and I believe we can avoid it).
</blockquote></details>

Done.
___
*[`drake_ros/core/ros_idl_pybind.h` line 113 at r9](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NYU_snt2PfkiquocnWu:-NYUgQqe1kGBgK-fqUZV:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/918db4b3c0567896b1d4f651530a5b6f631c0a8a/drake_ros/core/ros_idl_pybind.h#L113)):*

nit These other using statemetns can be placed in a macro.
</blockquote></details>

Done.
___
*[`drake_ros/drake_ros/cc_py_generic_typecaster.cc` line 14 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrjWgw0cby4oM0Pr2t:-NYUgTnC3Guh_qYJKSNn:b-lzq7il) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/drake_ros/cc_py_generic_typecaster.cc#L14)):*

Yeah, I'm finding the same behavior. In this case, we should just create separate modules.
</blockquote></details>

Done.

My guess is that the linker was able to find specializations of `type_caster<T>`... though I'm not exactly sure how that works.

EricCousineau-TRI (commented):
:lgtm: w/ adjustments and pending CI, thanks!

> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-:-NYUh84U1cEfWSxLsIvs:b-371j9j)* status: 2 of 12 files reviewed, all discussions resolved (waiting on @sloretz)

*[a discussion](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NYU_48Q-6nw0Z0x0ds-:-NYUh8yF-4XEhwbpwQZD:b-896fix) (no related file):*

Working: I have a few other fixes incoming.
</blockquote></details>

Done.

___
___
*[`drake_ros/core/geometry_conversions_pybind.h` line 8 at r7](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NSrk2mN9gE3YJ0YU2eq:-NYUh49k1olV7doxsj_O:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/ca5400a06c402e09862360b050e340c22248ec97/drake_ros/core/geometry_conversions_pybind.h#L8)):*

Working: Will fix.
</blockquote></details>

Done.
___
*[`.gitignore` line 2 at r10](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NYUgMub79c8yNqvRFyf:-NYUh5eHDqAJuWZ_STd_:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/d1c7fad2d29c6c0962235d0f17b49fc5a329427a/.gitignore#L2)):*

Working: Oops should remove this.
</blockquote></details>

Done.
___
*[`drake_ros/drake_ros/test/ros_message_type_caster_test_via_specific_types_pybind.cc` line 36 at r9](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257#-NYUVOCM6wsn9sJlQjuU:-NYUh6s17Q921YUeF3dp:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/918db4b3c0567896b1d4f651530a5b6f631c0a8a/drake_ros/drake_ros/test/ros_message_type_caster_test_via_specific_types_pybind.cc#L36)):*

nit Unused?
</blockquote></details>

Done.

EricCousineau-TRI (approved):
> Reviewed 6 of 9 files at r10, 3 of 3 files at r11, 1 of 2 files at r12, 1 of 1 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/257)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @adityapande-1995)

PR #284, opened 2023-06-13
Workaround ros2/rosidl_typesupport#120

sloretz:
This works around ros2/rosidl_typesupport#120.

When the ROS workspace has multiple typesupports, and the RMW implementation uses the introspection typesupport, and `rosidl_interfaces_group()` is given given a value for `group`, the introspection typesupport fails to load language specific shared libraries at runtime.

This works around the bug by making the generated target names use the group name as their prefix, and then making `alias()` targets with `name` used as the prefix. This makes sure the generated shared libraries have the name that the introspection type support expects.

If the upstream issue is fixed, this workaround will continue to work.

@IanTheEngineer FYI

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/284)

### DIFF
--- a/bazel_ros2_rules/ros2/rosidl.bzl
+++ b/bazel_ros2_rules/ros2/rosidl.bzl
@@ -1316,6 +1316,18 @@ def rosidl_interfaces_group(
 
     Additional keyword arguments are those common to all rules.
     """
+
+    # Workaround ros2/rosidl_typesupport#120
+    # The introspection type supports assume the library name is the same as
+    # the package name (aka "group" here). If the ROS workspace supports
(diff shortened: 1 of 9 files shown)


IanTheEngineer (approved):
:LGTM: This workaround looks reasonable to me, and passes our CI.

> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/284#-:-NYPMgUSBaZW0tXpIBZJ:bp8j1za)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sloretz)

Issue #247, opened 2023-03-07
[ci] Should improve bazel caching


EricCousineau-TRI:
The past few PRs have been very slow to build / use `drake@v1.13.0` per recent update.

May possible be improved by following up on https://drakedevelopers.slack.com/archives/C2CP2E50A/p1673880604846249

jwnimmer-tri:
I'm pretty sure that the caching is all wrong.

By default, GitHub Actions only _write back_ to the cache when the `key:` string changes.  Since our key string is a literal with no substitutions, it is never written back, so it's getting progressively colder as we continue to change the code.

See https://github.com/RobotLocomotion/drake-blender/pull/7 where I'm iterating to get this right.  I'll circle back here once I've tested it there on a smaller repo.

EricCousineau-TRI:
Ah, good to know, and thanks!

EricCousineau-TRI:
Getting close, but the job can still get cancelled sometimes:
https://github.com/RobotLocomotion/drake-ros/actions/runs/6303897969/job/17114097008

It seems like it is still running out of disk space:
https://github.com/RobotLocomotion/drake-ros/actions/runs/6303897969/attempts/1
```
Unhandled exception. System.IO.IOException: No space left on device : '/home/runner/runners/2.309.0/_diag/Worker_20230925-192117-utc.log'
```
@adityapande-1995 Would you be able to look into this?
\cc @calderpg-tri @IanTheEngineer 

FWIW, GitHub Workflows seems a bit wonky right now:
- I don't always see instructions for the current executing step
- If I re-run the job, then I cannot see the old logs

jwnimmer-tri:
BTW If you're running `apt` commands in the build (either update and/or upgrade), make sure you nuke their downloaded files and lists database afterwards:
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get

EricCousineau-TRI:
Resolved by #309

PR #290, opened 2023-08-07
[fix] Multirobot cpp version doesn't create the correct plant setup and viz 

adeeb10abbas:
As of now, currently the `multirobot` (cpp) version creates a 4x4 grid as compared to what it should be, a 5x5 one. The Python version works as expected. I have attached the videos which show the fewer robots + amputated kukas. 

Working Python version - 

https://github.com/RobotLocomotion/drake-ros/assets/38449494/7bab1543-740d-4146-9b44-b91d2e0db967

Broken Cpp version- 

https://github.com/RobotLocomotion/drake-ros/assets/38449494/f08671f0-83de-49b4-9d35-e33ffa8f963e

My guess is that `uint8_t` being compared with `size_t` seems to be the culprit. Although ideally it shouldn't, however, maybe it's some compiler settings that makes the treatment that ends up in this error?

cc: @EricCousineau-TRI for review, thanks!

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/290)

### DIFF
--- a/drake_ros_examples/examples/multirobot/multirobot.cc
+++ b/drake_ros_examples/examples/multirobot/multirobot.cc
@@ -74,12 +74,12 @@ int main(int argc, char** argv) {
   const std::string model_name = "kuka_iiwa";
 
   // Create a 5x5 array of manipulators
-  size_t kNumRows = 5;
-  size_t kNumCols = 5;
+  int kNumRows = 5;
+  int kNumCols = 5;
   std::vector<std::vector<drake::multibody::ModelInstanceIndex>> models;
-  for (uint8_t xx = 0; xx < kNumRows; ++xx) {
+  for (int xx = 0; xx < kNumRows; ++xx) {
     std::vector<drake::multibody::ModelInstanceIndex> models_xx;
-    for (uint8_t yy = 0; yy < kNumCols; ++yy) {
+    for (int yy = 0; yy < kNumCols; ++yy) {
       // Load the model from the file and give it a name based on its X and Y
       // coordinates in the array
       std::stringstream model_instance_name;
@@ -102,8 +102,8 @@ int main(int argc, char** argv) {
   plant.Finalize();
 
   // Set the control input of each robot to uncontrolled
-  for (size_t xx = 0; xx < kNumRows; ++xx) {
-    for (size_t yy = 0; yy < kNumCols; ++yy) {
+  for (int xx = 0; xx < kNumRows; ++xx) {
+    for (int yy = 0; yy < kNumCols; ++yy) {
       // Get the number of degrees of freedom for the robot
       auto num_dofs = plant.num_actuated_dofs(models[xx][yy]);
       // Create a vector with the same number of zeros


jwnimmer-tri:
An `operator<<` on `uint8_t` might be streaming it as a character, instead of an integer.

This is one reason why [Drake's C++ style guide](https://drake.mit.edu/styleguide/cppguide.html#Integer_Types) forbids most uses and `size_t` and especially forbids using `uint8_t` for numbers.

EricCousineau-TRI:
Interesting... On my machine, using latest `main` (2d252c3), both the C++ and Python version look the same, so maybe it's an ARM thing?

Regardless, (a) I missed the usage of `uint8_t` in review (shouldn't have needed that) and (b) Jeremy's right, we should use `int` instead per the style guide.

EricCousineau-TRI (approved):
:lgtm: with a nit to use `int`

> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/290#-:-NbHSFyn34Pdoi4skKCo:b-15wxj0)* status: all files reviewed, 1 unresolved discussion (waiting on @adeeb10abbas)

___
*[`drake_ros_examples/examples/multirobot/multirobot.cc` line 80 at r1](https://reviewable.io/reviews/RobotLocomotion/drake-ros/290#-NbHRG7u4XYNWAnxyiLG:-NbHRG7u4XYNWAnxyiLH:b-j2mavu) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/c7397ec6df10dd5010a82f003777be8c4793c5e1/drake_ros_examples/examples/multirobot/multirobot.cc#L80)):*
> ```cpp
>   size_t kNumCols = 5;
>   std::vector<std::vector<drake::multibody::ModelInstanceIndex>> models;
>   for (size_t xx = 0; xx < kNumRows; ++xx) {
> ```

nit In this file, would be OK with replacing `size_t` with `int`?

adeeb10abbas:
> Interesting... On my machine, using latest `main` ([2d252c3](https://github.com/RobotLocomotion/drake-ros/commit/2d252c367f82c77fb74ce87a74c1dde73dfd8946)), both the C++ and Python version look the same, so maybe it's an ARM thing?

I am running `x86` as well. Jammy with an Intel chip. 

Yep! Jeremy's comments clear up the confusion. On a related note, why don't we bring the same drake linting and styleguide here? I think it would probably save a lot of time in the future as the codebase expands...

EricCousineau-TRI:
We bring some of the Drake related linting, and it'd be good to bring in other tooling, but I don't think I'd be able to do it anytime in near future. FYI @calderpg-tri @IanTheEngineer if y'all want to queue this up.

As far as style guide goes, we should do better in enforcing it, and it'd be good to eventually have platform review. Not sure on timeline there, though.

EricCousineau-TRI (approved):
:lgtm:, thanks for the changes!

> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/290#-:-NbHZEPT8knDUKVx3iwI:b-egq756)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @adeeb10abbas)

EricCousineau-TRI:
I am assuming CI failure is due to flake, possibly improved by #247 (\cc @IanTheEngineer @calderpg-tri)
Going to merge.

PR #306, opened 2023-10-27
[ci] Always save cache, even on failure

EricCousineau-TRI:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/306)


EricCousineau-TRI:
Got a partial cache, gonna move forward with this.

PR #350, opened 2024-05-13
[wip] rviz_default_plugins breaks on some systems

EricCousineau-TRI:
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/350)


EricCousineau-TRI:
per cb9c9db, it seems like this happened due to partial upgrade, done while bisecting to discovery https://github.com/ros2/geometry2/issues/676

closing as "not a problem for now"

PR #359, opened 2024-07-09
remove extra ; that cause a warning

TomCC7:
error message:
```bash
--- stderr: drake_ros
/workspace/codes/cc_control_ws/src/drake-ros/drake_ros/viz/scene_markers_system.cc:39:58: error: extra ‘;’ [-Werror=pedantic]
   39 |   DRAKE_NO_COPY_NO_MOVE_NO_ASSIGN(SceneGeometryToMarkers);
      |                                                          ^
      |                                                          -
cc1plus: all warnings being treated as errors
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/359)

### DIFF
--- a/drake_ros/viz/scene_markers_system.cc
+++ b/drake_ros/viz/scene_markers_system.cc
@@ -36,7 +36,7 @@ namespace {
 /// Converts Drake shape descriptions to ROS Marker messages.
 class SceneGeometryToMarkers : public drake::geometry::ShapeReifier {
  public:
-  DRAKE_NO_COPY_NO_MOVE_NO_ASSIGN(SceneGeometryToMarkers);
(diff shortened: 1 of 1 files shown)


jwnimmer-tri:
Apologies for the trouble, and thanks for bringing to our attention.

The lack of a semicolon is deprecated in the current Drake nightlies, and the semicolon will be strictly required in a future Drake release.  So we need to keep the semicolon there.

Instead, I've filed #360 which hopefully fixes the error.  Turning on `-Werror=pedantic` is aggressively disruptive, so the build should never have done that in the first place.

jwnimmer-tri:
Closing in lieu of #360.  If that PR doesn't actually fix your problem, just let us know.

PR #419, opened 2026-04-09
Improve ros_launch support for launch files

### LINKED ISSUE #417: ros_launch has no first-class support for Bazel-built nodes
#344 describes this as a known gap. This issue provides a concrete MRE and documents the full set of obstacles.

The natural ROS 2 pattern for launching a node:

```python
launch_ros.actions.Node(package="my_pkg", executable="my_node")
```
does not work for Bazel-built nodes. They are never installed into an ament prefix, so ROS 2's libexec lookup (`<prefix>/lib/my_pkg/my_node`) always fails.

## What doesn't work
```python
import launch_ros.actions
from launch import LaunchDescription


def generate_launch_description():

    # This is how you'd naturally write it -- but it fails under Bazel because
    # `eg_talker` and `eg_listener` are not installed in any ament prefix.
    return LaunchDescription(
        [
            launch_ros.actions.Node(
                package="ros2_example_apps",
                executable="eg_talker",
            ),
            launch_ros.actions.Node(
                package="ros2_example_apps",
                executable="eg_listener",
            ),
        ]
    )
```

```
ros_launch(
    name = "roslaunch_eg_node_problem",
    data = [
        ":eg_listener",
        ":eg_talker",
    ],
    launch_file = "eg_launch_node_problem.py",
    workspace_name = ros2_example_bazel_installed,
)
```

## A workaround

```
import launch_ros.actions
from launch import LaunchDescription
from python.runfiles import runfiles


def generate_launch_description():
    r = runfiles.Create()
    prefix = "ros2_example_bazel_installed/ros2_example_apps"

    return LaunchDescription(
        [
            launch_ros.actions.Node(
                executable=r.Rlocation(f"{prefix}/eg_talker", source_repo=""),
            ),
            launch_ros.actions.Node(
                executable=r.Rlocation(f"{prefix}/eg_listener", source_repo=""),
            ),
        ]
    )
```

## What it would be good to have

1. No hardcoded runfiles path strings in launch files
2. Build-time validation that referenced executables exist
3. `source_repo=""` handled transparently, not by the user

Also, the `source_repo` requirement is a current bug of the current launch file support:

```
bazel run //ros2_example_apps:roslaunch_eg_py
.
.
.
[INFO] [launch]: All log files can be found below /home/frn/.ros/log/2026-04-07-15-28-43-711973-vader-2564850
[INFO] [launch]: Default logging verbosity is set to INFO
[ERROR] [launch]: Caught exception in launch (see debug for traceback): Caught multiple exceptions when trying to load file of format [py]:
 - ValueError: /home/frn/drake-ros/ros2_example_bazel_installed/ros2_example_apps/eg_launch.py does not lie under the runfiles root /home/frn/.cache/bazel/_bazel_frn/a779afcb46037298559646c7c2f921d4/execroot/_main/bazel-out/k8-fastbuild/bin/ros2_example_apps/roslaunch_eg_py.runfiles
 - InvalidFrontendLaunchFileError: The launch file may have a syntax error, or its format is unknown
```

julianadrianheine:
# Motivation
Previously, `ros_launch` required users to manually resolve Bazel runfile paths and use `ExecuteProcess` to launch nodes. This PR enables the standard ROS 2 launch pattern (`launch_ros.actions.Node`) to work with Bazel-built binaries by registering them in a fake ament prefix at build time.

For more details, you can check out this issue: https://github.com/RobotLocomotion/drake-ros/issues/417

# Implementation details
- Extends `ament_index_share_files` in `ament_index.bzl` to support an `executables` attribute, which symlinks Bazel-built binaries under `<prefix>/lib/<package_name>/`. This makes them discoverable by `launch_ros.actions.Node(package=..., executable=...)` without a colcon install space.
- Adds an `executables` parameter to ros_launch: a dict mapping ROS 2 package names to lists of executable targets. The macro automatically creates the ament index entries and wires up the runfiles.
- Removes the `data` parameter from `ros_launch` (executables registered via executables are automatically added to the binary's runfiles).
- Updates the example launch app to use the idiomatic `launch_ros.actions.Node` pattern instead of resolving and spawning binaries manually via `ExecuteProcess` + runfiles.

- - -
This change is [Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419)

### DIFF
--- a/bazel_ros2_rules/lib/ament_index.bzl
+++ b/bazel_ros2_rules/lib/ament_index.bzl
@@ -73,17 +73,14 @@ def _ament_index_share_files_impl(ctx):
             )
             root_symlinks[symlink_path] = file
 
-    for exe in ctx.attr.executables:
-        exe_file = exe.files_to_run.executable
-        if exe_file == None:
-            continue
+    for executable in ctx.attr.executables:
         symlink_path = paths.join(
             ctx.attr.prefix,
             "lib",
             ctx.attr.package_name,
-            exe_file.basename,
+            executable.files_to_run.executable.basename,
         )
-        root_symlinks[symlink_path] = exe_file
+        root_symlinks[symlink_path] = executable.files_to_run.executable
 
     return [
         AmentIndex(prefix = ctx.attr.prefix),
@@ -101,7 +98,7 @@ ament_index_share_files = rule(
         ),
         executables = attr.label_list(
             allow_empty = True,
-            allow_files = True,
+            allow_files = False,
         ),
         # A prefix is required because the shim can't prepend the runfiles
         # root to AMENT_PREFIX_PATH
--- a/bazel_ros2_rules/lib/ament_index.bzl
+++ b/bazel_ros2_rules/lib/ament_index.bzl
@@ -74,6 +74,8 @@ def _ament_index_share_files_impl(ctx):
             root_symlinks[symlink_path] = file
 
     for executable in ctx.attr.executables:
+        if executable.files_to_run.executable == None:
+            fail("{} is not an executable target".format(executable.label))
         symlink_path = paths.join(
             ctx.attr.prefix,
             "lib",
--- a/bazel_ros2_rules/lib/ament_index.bzl
+++ b/bazel_ros2_rules/lib/ament_index.bzl
@@ -47,22 +47,15 @@ Recursively aggregates AmentIndex prefixes from dependencies into one provider.
 """
 
 def _ament_index_share_files_impl(ctx):
-    # declare that a "package" with the given name exists
     package_marker_path = paths.join(
         ctx.attr.prefix,
         "share/ament_index/resource_index/packages/",
         ctx.attr.package_name,
     )
     package_marker_out = ctx.actions.declare_file(package_marker_path)
-    ctx.actions.write(
-        output = package_marker_out,
-        content = "",
-    )
+    ctx.actions.write(output = package_marker_out, content = "")
 
-    # Symlink sources into the share directory
-    runfiles_symlinks = {
-        package_marker_path: package_marker_out,
-    }
+    root_symlinks = {package_marker_path: package_marker_out}
 
     for src in ctx.attr.srcs:
         for file in depset(transitive = [
@@ -78,12 +71,24 @@ def _ament_index_share_files_impl(ctx):
                 ctx.attr.package_name,
                 sp,
             )
-            runfiles_symlinks[symlink_path] = file
+            root_symlinks[symlink_path] = file
+
+    for exe in ctx.attr.executables:
+        exe_file = exe.files_to_run.executable
+        if exe_file == None:
+            continue
+        symlink_path = paths.join(
+            ctx.attr.prefix,
+            "lib",
+            ctx.attr.package_name,
+            exe_file.basename,
+        )
+        root_symlinks[symlink_path] = exe_file
 
     return [
         AmentIndex(prefix = ctx.attr.prefix),
         DefaultInfo(
-            runfiles = ctx.runfiles(root_symlinks = runfiles_symlinks),
+            runfiles = ctx.runfiles(root_symlinks = root_symlinks),
         ),
     ]
 
@@ -91,8 +96,11 @@ ament_index_share_files = rule(
     attrs = dict(
         package_name = attr.string(mandatory = True),
         srcs = attr.label_list(
-            mandatory = True,
-            allow_empty = False,
+            allow_empty = True,
+            allow_files = True,
+        ),
+        executables = attr.label_list(
+            allow_empty = True,
             allow_files = True,
         ),
         # A prefix is required because the shim can't prepend the runfiles
@@ -105,14 +113,17 @@ ament_index_share_files = rule(
     provides = [AmentIndex],
 )
 """
-Creates an ament resource index and share/ directory for a single package.
+Creates an ament resource index for a single ROS 2 package.
 
-This creates both a package marker for a package, and a share directory
-for that package.
-Files in `srcs` will be added to `share/package_name`.
-A target depending on this one will be able to access the files in the
-share directory using the package's share directory joined with the
-files they expect.
+Files in `srcs` are symlinked under `<prefix>/share/<package_name>/`,
+enabling ament_index lookups such as `get_package_share_directory()`.
+
+Executable targets in `executables` are symlinked under
+`<prefix>/lib/<package_name>/`, enabling
+`launch_ros.actions.Node(package=..., executable=...)` to find Bazel-built
+binaries without a colcon install space.
+
+Both can be used together in a single rule invocation.
 
     d = ament_index_cpp::get_package_share_directory("package_name")
     file = join(d, "short_path of file")
@@ -132,11 +143,12 @@ TODO(sloretz) detect and error when a target is given two ament indexes
 with the same package.
 
 Args:
-    package_name: name of a ROS 2 package to which these share files belong
+    package_name: name of a ROS 2 package
     srcs: targets whose files are placed into the share directory. Both
       direct files and transitive runfiles are included.
+    executables: executable targets to expose under lib/.
     prefix: optional prefix to give to the generated runfiles.
-    strip_prefix: optional prefix to strip from the short_path of the files
+    strip_prefix: optional prefix to strip from the short_path of srcs files.
 
 Provides:
     AmentIndex: a prefix path where the ament resource index was generated.
--- a/bazel_ros2_rules/lib/private/ros_py.bzl
+++ b/bazel_ros2_rules/lib/private/ros_py.bzl
@@ -195,7 +195,6 @@ def ros_launch(
         name,
         launch_file,
         args = [],
-        data = [],
         deps = [],
         visibility = None,
         # TODO(eric.cousineau): Remove this once Bazel provides a way to tell
@@ -213,6 +212,7 @@ def ros_launch(
         #   }
         executables = {},
         **kwargs):
+    data = []
     main = "{}_roslaunch_main.py".format(name)
     launch_respath = _make_respath(launch_file, workspace_name)
 
@@ -234,6 +234,8 @@ def ros_launch(
     )
 
     for pkg_name, pkg_executables in executables.items():
+        if type(pkg_executables) != type([]):
+            fail("executables['{}'] must be a list, got {}".format(pkg_name, type(pkg_executables)))
         index_target = "_{}_ament_index_{}".format(name, pkg_name)
         ament_index_share_files(
             name = index_target,
--- a/bazel_ros2_rules/lib/private/ros_py.bzl
+++ b/bazel_ros2_rules/lib/private/ros_py.bzl
@@ -1,6 +1,10 @@
 # -*- python -*-
 
 load(
+    "@bazel_ros2_rules//lib:ament_index.bzl",
+    "ament_index_share_files",
+)
+load(
     "@bazel_ros2_rules//lib:kwargs.bzl",
     "filter_to_only_common_kwargs",
     "remove_test_specific_kwargs",
@@ -198,6 +202,16 @@ def ros_launch(
         # runfiles.py to use "this repository" in a way that doesn't require
         # bespoke information.
         workspace_name = None,
+        # dict mapping ROS 2 package name to a list of executable
+        # targets to expose via the ament resource index, enabling
+        # launch_ros.actions.Node(package=..., executable=...) to find
+        # Bazel-built binaries without a colcon install space.
+        # Example:
+        #   executables = {
+        #       "my_pkg": [":talker", ":listener"],
+        #       "other_pkg": ["//other_pkg:some_node"],
+        #   }
+        executables = {},
         **kwargs):
     main = "{}_roslaunch_main.py".format(name)
     launch_respath = _make_respath(launch_file, workspace_name)
@@ -218,6 +232,17 @@ def ros_launch(
             REPOSITORY_ROOT + ":ros2",
         ],
     )
+
+    for pkg_name, pkg_executables in executables.items():
+        index_target = "_{}_ament_index_{}".format(name, pkg_name)
+        ament_index_share_files(
+            name = index_target,
+            package_name = pkg_name,
+            executables = pkg_executables,
+            visibility = ["//visibility:private"],
+        )
+        data = data + pkg_executables + [":" + index_target]
+
     data = data + [launch_file]
 
     if "tags" not in kwargs:
--- a/bazel_ros2_rules/lib/private/ros_py.bzl
+++ b/bazel_ros2_rules/lib/private/ros_py.bzl
@@ -178,12 +178,7 @@ os.execv(ros2_bin, args)
 def _make_respath(relpath, workspace_name):
     repo = native.repository_name()
     if repo == "@":
-        if workspace_name == None:
-            fail(
-                "Please provide `ros_launch(*, workspace_name)` so that " +
-                "paths can be resolved properly",
-            )
-        repo = workspace_name
+        repo = workspace_name if workspace_name != None else native.module_name()
     pkg = native.package_name()
     if pkg != "":
         pieces = [repo, pkg, relpath]
@@ -195,24 +190,26 @@ def ros_launch(
         name,
         launch_file,
         args = [],
+        data = [],
         deps = [],
         visibility = None,
-        # TODO(eric.cousineau): Remove this once Bazel provides a way to tell
-        # runfiles.py to use "this repository" in a way that doesn't require
-        # bespoke information.
+        # Optional: overrides the Bzlmod module name as the ament package name.
+        # Only needed when the desired package name differs from the module name.
         workspace_name = None,
-        # dict mapping ROS 2 package name to a list of executable
-        # targets to expose via the ament resource index, enabling
-        # launch_ros.actions.Node(package=..., executable=...) to find
-        # Bazel-built binaries without a colcon install space.
+        # Executable targets to expose via the ament resource index under
+        # lib/<package_name>/, enabling
+        # launch_ros.actions.Node(package=<package_name>, executable=...) to
+        # find Bazel-built binaries without a colcon install space.
         # Example:
-        #   executables = {
-        #       "my_pkg": [":talker", ":listener"],
-        #       "other_pkg": ["//other_pkg:some_node"],
-        #   }
-        executables = {},
+        #   executables = [":talker", ":listener"],
+        executables = [],
+        # Data file targets to expose via the ament resource index under
+        # share/<package_name>/, enabling FindPackageShare(<package_name>)
+        # and get_package_share_directory(<package_name>) to locate them.
+        # Example:
+        #   share = ["//my_pkg:config_files"],
+        share = [],
         **kwargs):
-    data = []
     main = "{}_roslaunch_main.py".format(name)
     launch_respath = _make_respath(launch_file, workspace_name)
 
@@ -233,17 +230,17 @@ def ros_launch(
         ],
     )
 
-    for pkg_name, pkg_executables in executables.items():
-        if type(pkg_executables) != type([]):
-            fail("executables['{}'] must be a list, got {}".format(pkg_name, type(pkg_executables)))
-        index_target = "_{}_ament_index_{}".format(name, pkg_name)
+    if executables or share:
+        package_name = workspace_name if workspace_name != None else native.module_name()
+        index_target = "_{}_ament_index".format(name)
         ament_index_share_files(
             name = index_target,
-            package_name = pkg_name,
-            executables = pkg_executables,
+            package_name = package_name,
+            executables = executables,
+            srcs = share,
             visibility = ["//visibility:private"],
         )
-        data = data + pkg_executables + [":" + index_target]
+        data = data + executables + share + [":" + index_target]
 
     data = data + [launch_file]
 
--- a/ros2_example_bazel_installed/ros2_example_apps/BUILD.bazel
+++ b/ros2_example_bazel_installed/ros2_example_apps/BUILD.bazel
@@ -240,12 +240,10 @@ workspace_name = "ros2_example_bazel_installed"
 # by registering the Bazel-built binaries in a fake ament prefix.
 ros_launch(
     name = "roslaunch_eg_py",
-    executables = {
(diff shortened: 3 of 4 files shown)


jwnimmer-tri:
FYI the package name is available as `native.package_name()` in macros.  Possibly that is less brittle than making the user type it in?

julianadrianheine:
> FYI the package name is available as `native.package_name()` in macros. Possibly that is less brittle than making the user type it in?

That's a really good point. I will address this one, thanks!

frneer (changes requested):
> @frneer reviewed 5 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419)* status: all files reviewed, 2 unresolved discussions (waiting on julianadrianheine).

___
*[`bazel_ros2_rules/lib/ament_index.bzl` line 160 at r1](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419#-OpnhW6-4-Atl-AxYpe6:-OpnhW6-4-Atl-AxYpe7:bh5kzs6) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/0a329485677aa6385d955943a1843a5151220450/bazel_ros2_rules/lib/ament_index.bzl#L160)):*
> ```Starlark
>     )
> 
> def ament_index_executables(name, package_name, srcs, prefix = "ament_index_share_files", **kwargs):
> ```

I'd suggest including the logic in the existing original rule. It seems pretty similar to me adding files to `share/` and to `/lib`.

Maybe the difference should be in how we are consuming the files from. Consider using a specific attribute for executables, instead of reading it from `data`.

```
ros_launch(
    name = "my_launch",
    launch_file = "my_launch.py",
    executables = [":my_node"], 
)
```

Another thing to consider related to this is, with the current format all listed nodes will be part of the package name passed to the `ros_launch` but that is not necessarily the case, we might be bringing nodes from other packages.

I've thought of having sth like:

```
ros_launch(
    name = "my_launch",
    launch_file = "my_launch.py",
    executables = {
          "package_1":my_node"
          "package_2":my_other_node",
    }      
)
```

So you can do:
```
           launch_ros.actions.Node(
                package="package_1",
                executable="my_node",
            ),
           launch_ros.actions.Node(
                package="package_2",
                executable="my_other_node",
            ),
```

But at the same time, we don't really care which package it came from, we only care about resolving the correct executable path without relying the RLocation... so maybe keeping a single package name is good enough.
___
*[`ros2_example_bazel_installed/ros2_example_apps/eg_launch.py` line 17 at r1](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419#-OpnmI2cERacoCMaoqF6:-OpnmI2cERacoCMaoqF7:b97z04a) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/0a329485677aa6385d955943a1843a5151220450/ros2_example_bazel_installed/ros2_example_apps/eg_launch.py#L17)):*
> ```Python
>         [
>             # Running a talker written in python.
>             ExecuteProcess(cmd=[talker_bin]),
> ```

this example should follow the new way of using `ros_launch`.

julianadrianheine (commented):
> @julianadrianheine made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on frneer).

___
*[`bazel_ros2_rules/lib/ament_index.bzl` line 160 at r1](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419#-OpnhW6-4-Atl-AxYpe6:-Ops5sMm2kyjzaWBsnWr:btxj9kw) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/0a329485677aa6385d955943a1843a5151220450/bazel_ros2_rules/lib/ament_index.bzl#L160)):*
<details><summary><i>Previously, frneer (Francisco Rossi) wrote…</i></summary><blockquote>

I'd suggest including the logic in the existing original rule. It seems pretty similar to me adding files to `share/` and to `/lib`.

Maybe the difference should be in how we are consuming the files from. Consider using a specific attribute for executables, instead of reading it from `data`.

```
ros_launch(
    name = "my_launch",
    launch_file = "my_launch.py",
    executables = [":my_node"], 
)
```

Another thing to consider related to this is, with the current format all listed nodes will be part of the package name passed to the `ros_launch` but that is not necessarily the case, we might be bringing nodes from other packages.

I've thought of having sth like:

```
ros_launch(
    name = "my_launch",
    launch_file = "my_launch.py",
    executables = {
          "package_1":my_node"
          "package_2":my_other_node",
    }      
)
```

So you can do:
```
           launch_ros.actions.Node(
                package="package_1",
                executable="my_node",
            ),
           launch_ros.actions.Node(
                package="package_2",
                executable="my_other_node",
            ),
```

But at the same time, we don't really care which package it came from, we only care about resolving the correct executable path without relying the RLocation... so maybe keeping a single package name is good enough.
</blockquote></details>

Done.

- Regarding the first suggestion: instead of introducing a separate `ament_index_executables` rule, I added an `executables` argument directly to `ament_index_share_files`. The same argument was also added to `ros_launch`, which calls `ament_index_share_files` internally when `executables` is provided.

- Regarding the second suggestion: the `executables` argument in `ros_launch` is now a dict mapping ROS 2 package names to lists of Bazel executable targets. This allows nodes from multiple packages to be registered in the ament resource index. It is worth noting that the package name key is not validated by Bazel, what actually gets built and linked is determined by the Bazel labels. The package name only needs to be consistent between the `ros_launch` call and the `package=` argument in the corresponding `launch_ros.actions.Node` call in the launch file. The conventional choice is to use the same package name as in a colcon build, which makes the launch file portable between both build systems without modification.
___
*[`ros2_example_bazel_installed/ros2_example_apps/eg_launch.py` line 17 at r1](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419#-OpnmI2cERacoCMaoqF6:-Ops97iS4u5bG-cUl3p1:b-rqko7i) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/0a329485677aa6385d955943a1843a5151220450/ros2_example_bazel_installed/ros2_example_apps/eg_launch.py#L17)):*
<details><summary><i>Previously, frneer (Francisco Rossi) wrote…</i></summary><blockquote>

this example should follow the new way of using `ros_launch`.
</blockquote></details>

Done.

Alright, I just wanted to add an extra example showing that both approaches are still possible. But I guess the usage of RLocation will be deprecated

frneer (changes requested):
Awesome work! Left some final commnets.

> @frneer reviewed 5 files and all commit messages, made 6 comments, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419#-:-Opt07B-Cw87N7Upxqu8:b-4mtm6b)* status: all files reviewed, 5 unresolved discussions (waiting on julianadrianheine).

___
*[`bazel_ros2_rules/lib/ament_index.bzl` line 101 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419#-Opswr2B3PKN7_8aDk9V:-Opswr2B3PKN7_8aDk9W:bkg616) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/9cdd9bfcf8436aedea2d9ab7f4ffb9879882a610/bazel_ros2_rules/lib/ament_index.bzl#L101)):*
Consider using `executable` attr label: https://bazel.build/rules/lib/toplevel/attr

> executable: [bool](https://bazel.build/rules/lib/core/bool.html); default is `False`  
True if the dependency has to be executable. This means the label must refer to an executable file, or to a rule that outputs an executable file. Access the label with `ctx.executable.<attribute_name>`.

Also `allow_files` should be left empty or set to false.



_Suggestion:_
```Starlark
        executables = attr.label_list(
            allow_empty = True,
            executable = True,
```
___
*[`bazel_ros2_rules/lib/ament_index.bzl` line 73 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419#-OpsvPxmBbseow2Lrv0v:-OpsvPxn5z9ZsQqav8iv:bi4pye6) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/9cdd9bfcf8436aedea2d9ab7f4ffb9879882a610/bazel_ros2_rules/lib/ament_index.bzl#L73)):*
> ```Starlark
>             root_symlinks[symlink_path] = file
> 
>     for exe in ctx.attr.executables:
> ```

nit: don't use acronyms if unnecessary, it improves clarity.



_Suggestion:_
```Starlark
    for executable in ctx.attr.executables:
```
___
*[`bazel_ros2_rules/lib/ament_index.bzl` line 83 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419#-OpsxPbeBgkOt-CvLE5h:-OpsxPbeBgkOt-CvLE5i:b-jouy61) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/9cdd9bfcf8436aedea2d9ab7f4ffb9879882a610/bazel_ros2_rules/lib/ament_index.bzl#L83)):*
> ```Starlark
>             exe_file.basename,
>         )
>         root_symlinks[symlink_path] = exe_file
> ```

After applying the suggestion below we won't need to manually check if target is an executable.



_Suggestion:_
```Starlark
    for exe in ctx.attr.executables:
        symlink_path = paths.join(
            ctx.attr.prefix,
            "lib",
            ctx.attr.package_name,
            exe_file.basename,
        )
        root_symlinks[symlink_path] = exe.files_to_run.executable
```
___
*[`bazel_ros2_rules/lib/private/ros_py.bzl` line 236 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419#-Opt-enoEoubGGBEgIEA:-Opt-enoEoubGGBEgIEB:bd29lhk) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/9cdd9bfcf8436aedea2d9ab7f4ffb9879882a610/bazel_ros2_rules/lib/private/ros_py.bzl#L236)):*
> ```Starlark
>     )
> 
>     for pkg_name, pkg_executables in executables.items():
> ```

nit: should we validate `pkg_executables` is in deed a list?
___
*[`bazel_ros2_rules/lib/private/ros_py.bzl` line 244 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419#-Opt-wJs8sVhoBWljBwl:-Opt-wJs8sVhoBWljBwm:b-mwy7b7) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/9cdd9bfcf8436aedea2d9ab7f4ffb9879882a610/bazel_ros2_rules/lib/private/ros_py.bzl#L244)):*
> ```Starlark
>             visibility = ["//visibility:private"],
>         )
>         data = data + pkg_executables + [":" + index_target]
> ```

nit: maybe data is no longer a good naming?

julianadrianheine (commented):
> @julianadrianheine made 5 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419)* status: all files reviewed, 4 unresolved discussions (waiting on frneer).

___
*[`bazel_ros2_rules/lib/ament_index.bzl` line 73 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419#-OpsvPxmBbseow2Lrv0v:-Opt2FRD3vS3jkM4jOZZ:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/9cdd9bfcf8436aedea2d9ab7f4ffb9879882a610/bazel_ros2_rules/lib/ament_index.bzl#L73)):*
<details><summary><i>Previously, frneer (Francisco Rossi) wrote…</i></summary><blockquote>

nit: don't use acronyms if unnecessary, it improves clarity.
</blockquote></details>

Done.
___
*[`bazel_ros2_rules/lib/ament_index.bzl` line 83 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419#-OpsxPbeBgkOt-CvLE5h:-Opt2VG55Q32eLxl1yV3:b-896fix) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/9cdd9bfcf8436aedea2d9ab7f4ffb9879882a610/bazel_ros2_rules/lib/ament_index.bzl#L83)):*
<details><summary><i>Previously, frneer (Francisco Rossi) wrote…</i></summary><blockquote>

After applying the suggestion below we won't need to manually check if target is an executable.
</blockquote></details>

Done.
___
*[`bazel_ros2_rules/lib/ament_index.bzl` line 101 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419#-Opswr2B3PKN7_8aDk9V:-Opt7efe8xoj5T5AdDCo:bc8h9te) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/9cdd9bfcf8436aedea2d9ab7f4ffb9879882a610/bazel_ros2_rules/lib/ament_index.bzl#L101)):*
<details><summary><i>Previously, frneer (Francisco Rossi) wrote…</i></summary><blockquote>

Consider using `executable` attr label: https://bazel.build/rules/lib/toplevel/attr

> executable: [bool](https://bazel.build/rules/lib/core/bool.html); default is `False`  
True if the dependency has to be executable. This means the label must refer to an executable file, or to a rule that outputs an executable file. Access the label with `ctx.executable.<attribute_name>`.

Also `allow_files` should be left empty or set to false.
</blockquote></details>

Executable is part of label, but not of label_list actually.
___
*[`bazel_ros2_rules/lib/private/ros_py.bzl` line 236 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419#-Opt-enoEoubGGBEgIEA:-Opt4Pba5WKdZ9GXOQs8:b-a0p6x7) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/9cdd9bfcf8436aedea2d9ab7f4ffb9879882a610/bazel_ros2_rules/lib/private/ros_py.bzl#L236)):*
<details><summary><i>Previously, frneer (Francisco Rossi) wrote…</i></summary><blockquote>

nit: should we validate `pkg_executables` is in deed a list?
</blockquote></details>

That makes sense. I will add a check here!
___
*[`bazel_ros2_rules/lib/private/ros_py.bzl` line 244 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419#-Opt-wJs8sVhoBWljBwl:-Opt4T6b7ANZLFLbsHZ5:bimwdus) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/9cdd9bfcf8436aedea2d9ab7f4ffb9879882a610/bazel_ros2_rules/lib/private/ros_py.bzl#L244)):*
<details><summary><i>Previously, frneer (Francisco Rossi) wrote…</i></summary><blockquote>

nit: maybe data is no longer a good naming?
</blockquote></details>

That's a good point. `data` was left as an argument because it was needed for the xml launch files. But this is actually not needed, we can still use `ros_launch` with xml files as long as we build the file as

```
<launch>
<executable cmd="<pkg_name>/<node_name>" output="screen"/>
</launch>
```

We still need to create a list inside `ros_launch` to pass to `py_binary`

frneer (approved):
:lgtm:

> @frneer reviewed 3 files and all commit messages, made 2 comments, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419#-:-OptY6PUEnOI7I0BYqLB:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on julianadrianheine).

___
*[`bazel_ros2_rules/lib/ament_index.bzl` line 101 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419#-Opswr2B3PKN7_8aDk9V:-OptWkY49t0nC-VsUeji:b-trpbq) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/9cdd9bfcf8436aedea2d9ab7f4ffb9879882a610/bazel_ros2_rules/lib/ament_index.bzl#L101)):*
<details><summary><i>Previously, julianadrianheine (julianheine) wrote…</i></summary><blockquote>

Executable is part of label, but not of label_list actually.
</blockquote></details>

Roger, so now we'll fail when trying to access the attribute in case it's an executable right?

julianadrianheine (commented):
> @julianadrianheine made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419)* status: all files reviewed, 1 unresolved discussion (waiting on frneer).

___
*[`bazel_ros2_rules/lib/ament_index.bzl` line 101 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419#-Opswr2B3PKN7_8aDk9V:-Oq5boHxA4BpjthdB3OQ:b-8h81ye) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/9cdd9bfcf8436aedea2d9ab7f4ffb9879882a610/bazel_ros2_rules/lib/ament_index.bzl#L101)):*
<details><summary><i>Previously, frneer (Francisco Rossi) wrote…</i></summary><blockquote>

Roger, so now we'll fail when trying to access the attribute in case it's an executable right?
</blockquote></details>

Sorry, I don't think I follow. Since we actually want these files to be executables, we shouldn't fail. Right?

frneer (changes requested):
> @frneer made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419)* status: all files reviewed, 1 unresolved discussion (waiting on julianadrianheine).

___
*[`bazel_ros2_rules/lib/ament_index.bzl` line 101 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419#-Opswr2B3PKN7_8aDk9V:-Oq5iX3-2gWIbtXHHspQ:bpe8buc) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/9cdd9bfcf8436aedea2d9ab7f4ffb9879882a610/bazel_ros2_rules/lib/ament_index.bzl#L101)):*
<details><summary><i>Previously, julianadrianheine (julianheine) wrote…</i></summary><blockquote>

Sorry, I don't think I follow. Since we actually want these files to be executables, we shouldn't fail. Right?
</blockquote></details>

If you pass a non-executable that's not a file that won't be catch by `allow_files=False`

When passing a for example a `cc_library` you'll end up with this error:

```
ERROR: /home/frn/drake-ros/ros2_example_bazel_installed/ros2_example_apps/BUILD.bazel:241:11: in ament_index_share_files rule //ros2_example_apps:_roslaunch_eg_py_ament_index_ros2_example_apps: 
Traceback (most recent call last):
	File "/home/frn/.cache/bazel/_bazel_frn/a779afcb46037298559646c7c2f921d4/external/bazel_ros2_rules+/lib/ament_index.bzl", line 78, column 47, in _ament_index_share_files_impl
		executable.files_to_run.executable.basename,
Error: 'NoneType' value has no field or method 'basename'
```

Which is not impossible to debug but I think we can have a better behavior here, maybe explicitly failing when the user passes a non-executable non-file target?

julianadrianheine (commented):
> @julianadrianheine made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419)* status: all files reviewed, 1 unresolved discussion (waiting on frneer).

___
*[`bazel_ros2_rules/lib/ament_index.bzl` line 101 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419#-Opswr2B3PKN7_8aDk9V:-Oq5k_5tBSFb9mZjze_N:b-ouqz6r) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/9cdd9bfcf8436aedea2d9ab7f4ffb9879882a610/bazel_ros2_rules/lib/ament_index.bzl#L101)):*
<details><summary><i>Previously, frneer (Francisco Rossi) wrote…</i></summary><blockquote>

If you pass a non-executable that's not a file that won't be catch by `allow_files=False`

When passing a for example a `cc_library` you'll end up with this error:

```
ERROR: /home/frn/drake-ros/ros2_example_bazel_installed/ros2_example_apps/BUILD.bazel:241:11: in ament_index_share_files rule //ros2_example_apps:_roslaunch_eg_py_ament_index_ros2_example_apps: 
Traceback (most recent call last):
	File "/home/frn/.cache/bazel/_bazel_frn/a779afcb46037298559646c7c2f921d4/external/bazel_ros2_rules+/lib/ament_index.bzl", line 78, column 47, in _ament_index_share_files_impl
		executable.files_to_run.executable.basename,
Error: 'NoneType' value has no field or method 'basename'
```

Which is not impossible to debug but I think we can have a better behavior here, maybe explicitly failing when the user passes a non-executable non-file target?
</blockquote></details>

Oh I get your point now, thanks! Yes, we can add a check to verify if `executable.files_to_run.executable == None`

Now, if passing for example a `cc_library` the error is more explicit:
```
Traceback (most recent call last):
	File "/home/julianheine/.cache/bazel/_bazel_julianheine/564c480f292c1ed4c1cadf898631f510/external/bazel_ros2_rules+/lib/ament_index.bzl", line 75, column 17, in _ament_index_share_files_impl
		fail("{} is not an executable target".format(executable.label))
Error in fail: @@//ros2_example_apps:listener_cc is not an executable target
```

frneer (approved):
> @frneer reviewed 1 file and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on julianadrianheine).

___
*[`bazel_ros2_rules/lib/ament_index.bzl` line 101 at r2](https://reviewable.io/reviews/RobotLocomotion/drake-ros/419#-Opswr2B3PKN7_8aDk9V:-Oq5pio6EU9f4Setlkfa:bccpe83) ([raw file](https://github.com/RobotLocomotion/drake-ros/blob/9cdd9bfcf8436aedea2d9ab7f4ffb9879882a610/bazel_ros2_rules/lib/ament_index.bzl#L101)):*
<details><summary><i>Previously, julianadrianheine (julianheine) wrote…</i></summary><blockquote>

Oh I get your point now, thanks! Yes, we can add a check to verify if `executable.files_to_run.executable == None`

Now, if passing for example a `cc_library` the error is more explicit:
```
Traceback (most recent call last):
	File "/home/julianheine/.cache/bazel/_bazel_julianheine/564c480f292c1ed4c1cadf898631f510/external/bazel_ros2_rules+/lib/ament_index.bzl", line 75, column 17, in _ament_index_share_files_impl
		fail("{} is not an executable target".format(executable.label))
Error in fail: @@//ros2_example_apps:listener_cc is not an executable target
```
</blockquote></details>

Awesome!

